PR #1, opened 2016-04-06
Minor tweaks

longhotsummer:
* updated dependencies (old ones didn't work for me)
* refactor cubes into cubes.py
* fix 404 for bad cube names

@jbothma


longhotsummer:
Also changed the default username and password so that it is closer to production, so it's easier to setup locally.

jbothma:
Nice. LGTM

PR #6, opened 2016-04-14
Initial docs

jbothma:
just get something out there that we can put on the event page and start getting feedback on

@longhotsummer


longhotsummer:
:+1:

PR #8, opened 2016-04-15
Handle API 500-level errors as JSON

longhotsummer:


longhotsummer:
@jbothma

jbothma:
Nice. 
We can improve on this with https://www.pivotaltracker.com/story/show/117621339 - I'm sure there's a specific exception we can look for for parser errors but this gives the most urgent feedback people need so this is good to have immediately

longhotsummer:
Agreed. This is explicitly for 500s.

PR #12, opened 2016-04-18
Refix pipeline issue

longhotsummer:
@jbothma 

Turns out that when people write bad javascript, bad things happen.


jbothma:
And when things aren't strict errors propagate to hide the root cause #javascript

PR #13, opened 2016-04-18
Consistent names

jbothma:
unifying all cubes to refer to item as item is important for automatically the items as in the return forms. Also time to clean up the names now we understand their usage better before they get used a lot and we break more things by changing them later.

@longhotsummer FYI


models/aged_creditor.json
```diff
@@ -7,21 +7,21 @@
             "attributes": {
                 "code": {
                     "column": "demarcation_code",
-                    "label": "Code",
+                    "label": "Demarcation Code",
                     "type": "string",
                     "description": "Demarcation Code desc"
```
longhotsummer: While you're at it, why not make this:

Municipality code assigned by the Municipal Demarcation Board

models/aged_creditor.json
```diff
@@ -7,21 +7,21 @@
...
                 },
                 "desc": {
-                    "column": "demarcation_desc",
-                    "label": "Description",
+                    "column": "demarcation_label",
+                    "label": "Name",
                     "type": "string",
                     "description": "some description"
```
longhotsummer: Municipality name

models/aged_creditor.json
```diff
@@ -7,21 +7,21 @@
...
                 }
             },
             "key_attribute": "code",
             "label_attribute": "desc",
             "description": "Demarcation of location or some such",
```
longhotsummer: Municipality demarcation details.

longhotsummer:
Let's take this chance to fill in some of the other placeholders, too.

models/aged_creditor.json
jbothma: Sure. I like. For codes I think we can say where they're defined - e.g. I'll say which codes are the treasury's. This one's the demarcation board's etc.

models/aged_creditor.json
jbothma: I think this gets redundant when the dimension label is municipality and the attribute label is Name. I think it's best leaving the description blank in such cases

longhotsummer: I'm happy with that.

jbothma:
This is so freaking tedious and repetitive. Only reason I haven't templated this is because this shouldn't need a lot of changes once it's properly released

PR #15, opened 2016-04-19
Cache expensive API calls on client and server.

longhotsummer:
@jbothma


jbothma:
nice! This is becoming important

PR #20, opened 2016-04-20
Put a basic indicator on the page

xybrnet:
This gets a couple of numbers from the API,
and prints the calculated ratio on the profile page.

@longhotsummer


longhotsummer:
Good starting point.

PR #23, opened 2016-04-21
Add nav tabs to change between cubes in explorer

jbothma:


longhotsummer:
ugly, but functional... :)

jbothma:
harsh :( but true :D 

You should see it rocking on mobile!

PR #27, opened 2016-04-22
Add item docs

jbothma:
looks ok when it's simple

items-simple

looks bad when it's not simple

items-too-busy

and we need to figure out what's up with the cash flow data - why we have codes that they haven't defined anywhere

items-cflow-missing

### DIFF
--- a/sql/labels.sql
+++ b/sql/labels.sql
@@ -1,0 +1,162 @@
+drop table bsheet_labels;
+drop table aged_creditors_labels;
+drop table aged_debtors_labels;
+drop table capital_labels;
+drop table cflow_labels;
+drop table incexp_labels;
+drop table repmaint_labels;
+
+CREATE TABLE public.bsheet_labels
+(
+  code text NOT NULL,
+  label text,
+  position_in_return_form integer,
+  return_form_structure text,
+  composition text,
+  CONSTRAINT bsheet_labels_primary_key PRIMARY KEY (code)
+)
+WITH (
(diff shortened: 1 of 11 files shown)


sql/labels.sql
```diff
@@ -0,0 +1,162 @@
...
+alter table incexp_labeled add column return_form_structure text;
+alter table incexp_labeled add column composition text;
+
+alter table repmaint_labeled add column position_in_return_form integer;
+alter table repmaint_labeled add column return_form_structure text;
+alter table repmaint_labeled add column composition text;
+
+update bsheet_labeled set position_in_return_form = subquery.position_in_return_form, return_form_structure = subquery.return_form_structure, composition = subquery.composition from (select * from bsheet_labels ) as subquery where subquery.code = item_code
+
+update aged_creditor_labeled set position_in_return_form = subquery.position_in_return_form, return_form_structure = subquery.return_form_structure, composition = subquery.composition from (select * from aged_creditors_labels ) as subquery where subquery.code = item_code;
```
longhotsummer: what do these lines do?

longhotsummer:
I think this is valuable. Let's put it in and make it better as we go.

sql/labels.sql
jbothma: This is stuff I manually run to update the tables. Right now I deploy by doing a dump locally and applying that to prod so this is more a reminder to recreate things. We should probably clean up the sql directory

PR #33, opened 2016-04-25
Separate apps for the scorecard and the api

longhotsummer:
Move the scorecard app into its own Django app, and use the sites framework to differentiate between them.


longhotsummer:
@xybrnet @jbothma

xybrnet:
Should profiles.py also move into scorecard?

longhotsummer:
Updated to move profiles.py and geo.py. Thanks @xybrnet

xybrnet:
👍

municipal_finance/settings.py
```diff
@@ -50,6 +52,11 @@
     'corsheaders',
 )
 
+# Sites
+if DEBUG:
+    SITE_ID = 2  # Scorecard
```
jbothma: where's this used?

longhotsummer: In production, the sites framework uses the domain of the incoming request to work out what site is active. In development, we don't have a domain and this setting tells Django which site to use. The ID is the id of the corresponding Site object in the database.

What, random integers in settings files aren't your thing? :)

municipal_finance/migrations/0001_initial.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.9.5 on 2016-04-25 11:05
+from __future__ import unicode_literals
+
+from django.db import migrations
+from django.contrib.sites.models import Site
+
+
+def create_sites(apps, schema_editor):
```
jbothma: did you write this manually or was this generated from something you did on the shell? I want to add municipal-finance.code4sa.org to either serve the API site or redirect to data.municipalmoney.org.za

PR #40, opened 2016-04-26
Initial capital budget

xybrnet:
I had to change the way the URL was constructed to get budgeted amounts.
@jbothma


longhotsummer:
nice, a pattern is starting to emerge here which we can use to clean this up a bit. good for now.👍

PR #42, opened 2016-04-27
Wider docs, tweaks to tables

longhotsummer:
@jbothma 

This is intended to replace #36

### DIFF
--- a/municipal_finance/static/stylesheets/docs.scss
+++ b/municipal_finance/static/stylesheets/docs.scss
@@ -11,12 +11,29 @@
   }
 }
 
-.subtotal-item {
-    font-weight: bold;
-    border-top: 1px solid black;
+table.items {
+    max-height: 500px;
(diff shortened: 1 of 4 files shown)


municipal_finance/static/stylesheets/docs.scss
```diff
@@ -11,12 +11,29 @@
...
 table.items {
-    max-height: 400px;
+    max-height: 500px;
     overflow-y: scroll;
-}
\ No newline at end of file
+
+    .heading-item {
+        font-weight: bold;
+        text-transform: uppercase;
```
jbothma: this made me think twice since there are already some headings that are caps and some that aren't in the same return forms and our data, but I think this makes it more readable. I think it's fine to make it uppercase and revert if we get feedback that it's actually confusing people. We have the item codes as the real identifier of a line item anyway.

jbothma:
nice

PR #47, opened 2016-04-28
Load geometry from mapit and ensure shapes are clickable

longhotsummer:
We're loading metros and local munis.


xybrnet:
Cool. I kind of get it 👍

PR #53, opened 2016-05-03
Add a page for exploring the table data

longhotsummer:
This lets a user explore table data in a more structured format, emphasising how it's collected.


jbothma:
Nice! I think this is a way better starting point to look at the data than babbage.

Are you thinking of linking to this from /docs?

- gorgeous progress bar
- would be nice if it remembered my muni selection when I change dataset
- aged creditor isn't in return form order. TOT is line nr 21. This is in item code order.
aged-creditor not in return form order
- would be nice to explain that the table looks funny because there isn't daa available for the query like so
empty results
- bit weird that it's truncated even though I can scroll it:
truncated_and_scrolling

longhotsummer:
> would be nice if it remembered my muni selection when I change dataset
Agreed, I'd like it to remember all query settings.

> aged creditor isn't in return form order. TOT is line nr 21. This is in item code order. 
Looks like that's wrong in the model, then.

PR #54, opened 2016-05-03
Historical values

xybrnet:


longhotsummer:
I see what you mean about this becoming awkward in the code.

scorecard/templates/profile/profile_detail.html
```diff
@@ -83,7 +83,7 @@ <h1 class="title">{{ geography.this.short_name }}</h1>
     <li>Results:</li>
     <ul>
     {% for year, result in cap_budget_diff.iteritems %}
-    <li>{{ year }}: {{ result }}</li>
+    <li>{{ year|stringformat:"i" }}: {{ result }}</li>
```
longhotsummer: What happens if we don't do this?

longhotsummer:
LGTM

PR #55, opened 2016-05-03
Jquery 1.12.3 since we still have to support IE 8

longhotsummer:
This also fixes a problem that javascript was included twice on the page.


jbothma:
This breaks the docs page for me
jquery-1 12

longhotsummer:
Try now? Although whether or not anything else works on IE8 is unclear. But we generally use the 1.x branch because of this. 8% of PMG users, for example, are on IE8.

jbothma:
Works in docs for me and was missing in babbage so added there too.

PR #59, opened 2016-05-03
Store the state of the table view

longhotsummer:
Store the table state in the querystring as it changes, and load it when the page first loads.

This requires changing the municipalities filter value to by IDs, rather than objects, so that they're easier to work with.


longhotsummer:
Merge #61 into this before merging this.

longhotsummer:
Ok, this can be merged when someone has had a look and is happy. @jbothma

municipal_finance/static/javascript/table.js
```diff
@@ -73,38 +91,60 @@
           return muni;
         });
 
+        // global municipalities list
         municipalities = _.indexBy(munis, 'demarcation_code');
+
+        // sanity check pre-loaded municipalities
+        self.filters.set('municipalities', _.select(self.filters.get('municipalities'), function(id) {
+          return !!municipalities[id];
```
jbothma: !!?

PR #61, opened 2016-05-03
Highlight items with particular codes mentioned in the query string

longhotsummer:
We don't persist these highlighted items.

Merge this before #59.

### DIFF
--- a/municipal_finance/static/javascript/table.js
+++ b/municipal_finance/static/javascript/table.js
@@ -397,6 +397,10 @@
       var cells = this.cells.get('items');
       var munis = this.filters.get('municipalities');
       var scale = Math.pow(10, Number.parseInt(this.scale));
+      // highlightable items as a set of codes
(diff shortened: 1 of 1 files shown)


municipal_finance/static/javascript/table.js
```diff
@@ -425,6 +432,13 @@
           }
         }
       }
+
+      // highlighted rows
+      for (var h = 0; h < toHighlight.length; h++) {
```
jbothma: bah - I wish [].forEach was supported in IE8

longhotsummer: Agreed. In some places I've used underscore's ``_.each(array, function(a) { ... } );`` but it's not really any more readable than a for loop and is harder to debug because of the function.

jbothma:
How do you use this? ?items=0100 doesn't work for me

PR #98, opened 2016-05-23
move audit opinion preprocessing code from python notebook to module

jbothma:


municipal_finance/data_import/audit_opinions.py
```diff
@@ -0,0 +1,58 @@
...
+"""
+import csv
+import pdb
+import sys
+import traceback
+import xlrd
+
+
+def convert(sheet, csv_file):
+    book = xlrd. open_workbook(sheet)
```
xybrnet: Is there a space here?

jbothma: there was - thanks for spotting it

PR #101, opened 2016-05-24
Remove wazicss

fritzjooste:


longhotsummer:
Pls make the equivalent change for the profile_detail template too.

longhotsummer:
@fritzjooste I've made a bunch of extra changes, can I merge this? It will probably collide with stuff you've done in master.

PR #103, opened 2016-05-25
Charts

xybrnet:


municipal_finance/static/javascript/charts.js
```diff
@@ -2,6 +2,8 @@ var Chart = function() {
   var self = this;
 
   self.init = function() {
+    self.thousands_format = d3.format(",.0f");
```
longhotsummer: SA uses a space as a thousands separator. Try this:

```
format = d3_format
  .formatLocale({decimal: ".", thousands: " ", grouping: [3], currency: "R"})
  .format(",.0f");
```

xybrnet:
I think these are good to start with.

PR #115, opened 2016-05-26
Use CachingJSONCubeManager

jbothma:
This reads each cube model from disk and introspects the DB table only once
per python process


longhotsummer:
LGTM

PR #116, opened 2016-05-26
Scorecard parallel api requests

jbothma:
https://github.com/ross/requests-futures

### DIFF
--- a/scorecard/profile_data.py
+++ b/scorecard/profile_data.py
@@ -1,4 +1,4 @@
-import requests
+from concurrent.futures import ThreadPoolExecutor
 from requests_futures.sessions import FuturesSession
 from collections import defaultdict, OrderedDict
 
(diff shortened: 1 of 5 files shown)


scorecard/profile_data.py
```diff
@@ -7,22 +8,22 @@
 
 class MuniApiClient(object):
     def __init__(self, geo_code):
-        self.API_URL = 'https://data.municipalmoney.org.za/api/cubes/'
```
longhotsummer: 👍

longhotsummer: I think this should be self.API_URL = settings.API_URL, rather than using settings inside the body of the class's methods.

scorecard/profile_data.py
```diff
@@ -7,22 +8,22 @@
...
-        if settings.DEBUG:
-            self.API_URL = 'http://127.0.0.1:8888/api/cubes/'
-
         self.geo_code = str(geo_code)
         self.line_item_params = self.get_line_item_params()
 
         self.results = defaultdict(dict)
         self.years = set()
+        responses = []
+        self.session = FuturesSession()
```
longhotsummer: If we create one MuniApiClient for every request, then this means we'll have a new thread pool for every request. That might be bad, causing a large number of threads. We may want to create a shared thread pool of up to X threads (for some sane X).

longhotsummer: Oh, is this what you meant by your commit comment?

longhotsummer:
I think we're now conflating the Scorecard calculator and the Muni API client. It's odd to me that the API Client does all its work inside a constructor, that seems like it would be hard to test or override behaviour. Would it be better to make the API client simpler and the scorecard calculator create an API client, then ask it to fetch details, rather than the API client magically knowing what data to fetch?

jbothma:
I don't think that conflation is new in this PR. I agree separating concerns is becoming important and the distinct concerns are becoming clear so that could be a good next step

jbothma:
This thread is adamant that the engine is threadsafe. I still can't conceive where the concurrent table introspection issue came from - it's not like python would be concurrently loading the cubes module and thus consructing engines. But anyway - can't reproduce now, and this code is logically better in terms of concurrency anyway. So let's call that ex-issue "can't reproduce".

longhotsummer:
Cool, LGTM

PR #117, opened 2016-05-27
Ui improvements from fritz

longhotsummer:


scorecard/templates/profile/profile_detail.html
```diff
@@ -66,39 +68,45 @@ <h1 class="title">{{ geography.this.short_name }}</h1>
...
+              <div class="col-sm-3">
+                <div class="person-info">
+                  <div class="person-role">{{ official.role }}</div>
+                  <div class="person-name"><i class="fa fa-fw fa-user" aria-hidden="true"></i> {{ official.title }} {{ official.name }}</div>
+                  <div class="person-contact-details">
+                    <a href="#" class="show-contact-details"><span class="icon"><i class="fa fa-plus" aria-hidden="true"></i></span> Contact Details</a>
+                    <!--
+                    <div class="contact-details">
+                      <ul class="fa-ul">
+                        <li>
```
jbothma: I tend to avoid merging commented out code. It's almost certainly out of date if it ever becomes prioritised to get merged - I rather keep it around in a well-named branch.

In this case, we don't want to show the number directly anyway - we want to guide them to contact the secretary anyway, right?

longhotsummer:
Lekker lekker. Feedback from the team is that the nav is much better. It really brings the performance up above the fold nicely.

I've made some minor tweaks (see the commits above), I'm merging.

scorecard/templates/profile/profile_detail.html
longhotsummer: In general I agree, in this case it was temporary, I've already replaced this with actual contact details and functioning "show contact details" buttons.

PR #119, opened 2016-05-27
Threadpool fixes

jbothma:


jbothma:
The ready hook is needed as __init__.py is imported before settings.py which depends on settings.py

PR #123, opened 2016-05-30
Report card errors

jbothma:


scorecard/profile_data.py
```diff
@@ -78,7 +78,7 @@ def aggregate_from_response(response, query_params, years):
         results = {}
         for code in query_params['cut']['item.code']:
           results[code] = OrderedDict([
-              (c['financial_period.period'], c[query_params['aggregate']])
+              (c['financial_period.period'], none_as_zero(c[query_params['aggregate']]))
```
longhotsummer: What about `c[query_params['aggregate']] or 0`?

jbothma: That was what I started with but felt it's a bit hidden. "where does the 0 come from? I'm sure the API returns a null" Do you think this function is making it more obscure?

longhotsummer: I think `or 0` and a comment would suffice. The function is overkill.

PR #135, opened 2016-05-31
Indicator calculation rows

xybrnet:


jbothma:
Nice

longhotsummer:
Nice, let's keep this open until Fritz can add some formatting.

It'd be nice if we can present the cube, code etc. so that when it's clicked it opens the table view.

PR #137, opened 2016-06-01
404 and 500 error pages

longhotsummer:
I was doing these for umibukela so copied them here quickly.

The 404 page pushes a Google Analytics event so that we can track them and their referrer.


jbothma:
Love the 404 trick

PR #149, opened 2016-06-02
Fix muni contacts

jbothma:
to deploy new urls:
- load new muni data with sql/table_and_data_scorecard_geography.sql

to deploy staff contacts table name change:
- rename table municipality_contacts to municipality_staff_contacts
- merge and push to dokku


jbothma:
my psql is giving a syntax error at the data copy stage with both the old and the new scorecard_geography script :/ I'm now on 9.5 - probably a version issue

longhotsummer:
What's the error?

jbothma:
Syntax error at 62 or whatever the first row of stdin is to the \copy
statement

longhotsummer:
I piped through sql/table_and_data_geography.sql without hassle. The municipality_staff_contacts table already exists on the server, I'm going to assume it has the correct data.

I've merged this to master and am pushing now.

PR #158, opened 2016-06-03
WIP Indicate when the contact details were last updated

jbothma:
One option is to have this in the model as done here. I don't think this justifies a cube. On the other hand this is a (cheap) query per cube.

The idea with last_updated is to be a partial ISO8061 time where the rest can be parsed as zeros - perhaps it's worth just having the full timestamp but that's kinda saying more than we can?

This is the precursor to being able to indicate somewhere "The contact details are correct as at xxxx and is updated quarterly"

The update_cycle values can match values in "period_length" which are currently year and month although "quarterly" and "annual" are really the terms used for this kind of thing

### DIFF
--- a/models/aged_creditor.json
+++ b/models/aged_creditor.json
@@ -2,6 +2,8 @@
     "fact_table": "aged_creditor_facts",
     "label": "Aged Creditor Analysis",
     "description": "This is based on the AC and ACA return forms.",
+    "update_cycle": "quarter",
+    "last_updated": "2015-06",
     "dimensions": {
         "demarcation": {
             "attributes": {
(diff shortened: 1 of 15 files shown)


jbothma:
@longhotsummer what do you think? About the priority of this and the solution. If you like this part of the solution and think it's a prio I'll do something for the frontend. I'm worried adding this to the scorecard makes it too much info but perhaps it's ok under contacts and for the rest it can be under "how calculated"

longhotsummer:
Sounds good to me. I'm happy with semi-static metadata on the model, I think it's a good approach.

longhotsummer:
Tweaked - @jbothma can I merge?

PR #160, opened 2016-06-03
Table view updates

longhotsummer:
* amount types are choosable
* make it work with more tables (Still breaks on some tables that don't have financial_period)
* remove scale
* download is a button in the top right, disable rather than hiding when it doesn't work


jbothma:
Very nice. For the other tables, a `model` query should tell you whether you wanna try and sort by position_in_return_form or item.code (when the former doesn't exist)

longhotsummer:
Yeah, I've folded that in. The only thing currently breaking is conditional grants because 'item' doesn't exist.

PR #169, opened 2016-06-07
Indicators for staff and contractor expenses

longhotsummer:
Also:

* draw charts based on data attributes to simplify things
* the data is injected automatically via the profileData var in javascript.

### DIFF
--- a/scorecard/profile_data.py
+++ b/scorecard/profile_data.py
@@ -466,25 +466,27 @@ class IndicatorCalculator(object):
         return values
 
     def expenditure_trends(self):
-        values = []
-
+        values = defaultdict(list)
         for year in self.years:
             try:
-                staff = self.results['expenditure_breakdown']['3000'][year] \
-                    + self.results['expenditure_breakdown']['3100'][year]
-                contracting = self.results['expenditure_breakdown']['4200'][year]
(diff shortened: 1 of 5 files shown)


longhotsummer:
@jbothma ok, ready to review and/or merge

longhotsummer:
Bleh, big change, sorry about that.

scorecard/profile_data.py
```diff
@@ -22,58 +21,361 @@
...
             }
         elif query['query_type'] == 'model':
             url = self.API_URL + query['cube'] + '/model'
             params = {}
-        return self.session.get(url, params=params, verify=False)
 
-    def response_to_results(self, api_response, query):
-        api_response.result().raise_for_status()
-        response_dict = api_response.result().json()
+        response = self.session.get(url, params=params, verify=False)
```
jbothma: I suspect calling get() and result() in the same loop will lose the benefit of futures. That's why we do one loop of get()s and another loop of result()s

longhotsummer: Oh, that makes sense. I didn't understand why it was done like that.

jbothma:
Nice. I don't think I caught all the changes but I like what's happening here

longhotsummer:
Yeah, a big pull request like this isn't helpful. My bad.

PR #174, opened 2016-06-07
Cash coverage lower bound

jbothma:


longhotsummer:
Heh, I've fixed the same calc docs in my expenses branch.

PR #175, opened 2016-06-07
Control where calculation items wrap

jbothma:
What do you think of this POC of controlling how the calculation items wrap? The natural wrapping is confusing. This helps a lot, I think.

Another issue visible below is that at this size the Show Calculation link is in a weird place compared to the calculations
wrap-control-and-calc-link


longhotsummer:
I like this. We could easily move the show calcs link below the row so that it's above the actual calcs.

PR #179, opened 2016-06-09
Print and PDF

longhotsummer:
* print-friendly
* pdf generation support

We still need to link in the PDF, and sort out icons


longhotsummer:
This is still a WIP

PR #185, opened 2016-06-10
Add simple link to incexp from income breakdown

jbothma:


scorecard/templates/profile/profile_detail.html
```diff
@@ -521,7 +521,8 @@ <h2 class="sub-heading">Cash Balance <span class="indicator-date">{{ latest.year
   <div class="indicator">
     <div class="row">
       <div class="col-sm-8">
-        <div class="chart-container narrow-margin" data-chart="grouped-bar-revenue_breakdown"></div>
+        <div class="chart-container narrow-margin" data-chart="grouped-bar-revenue_breakdown.values"></div>
```
longhotsummer: What's going on here? How did this work before hand if this is a fix?

jbothma: It's a fail, not a fix. I'm rushing. Thanks for catching.
This should have been backtracked from a solution I'm not using.

jbothma:
@longhotsummer fixed and **force-pushed**

PR #187, opened 2016-06-10
Explain badexp delay

jbothma:


scorecard/templates/profile/profile_detail.html
```diff
@@ -527,7 +527,8 @@ <h2 class="sub-heading">Cash Balance <span class="indicator-date">{{ latest.year
         <div class="panel panel-didyouknow">
           <div class="panel-heading"><i class="fa fa-info-circle" aria-hidden="true"></i> Did you know?</div>
           <div class="panel-body">
-            Fruitless and wasteful expenditure concerns spending which was made in vain and would have been avoided had reasonable care been exercised. An example of such expenditure would be paying a deposit for a venue and not using it, and so losing the deposit.
+            <p>Fruitless and wasteful expenditure concerns spending which was made in vain and would have been avoided had reasonable care been exercised. An example of such expenditure would be paying a deposit for a venue and not using it, and so losing the deposit.</p>
+            <p>Since calling expenditures fruitless or wasteful can involve quite a lot of debate, the numbers used are Restated audited amounts 18 months after the Financial Year End - part of the Medium Term Revenue and Expenditure Framework.</p>
```
longhotsummer: "... are the restated audit amounts issued 18 months after the financial year end... "

(capitalisation and grammar)

jbothma: I capitalised Restated because that seems to be an important term and I want people to think about that even if they won't be able to find out immediately what that means precisely. Do you think it's ok to capitalise it for that reason?

jbothma: Same with Financial Year End

PR #195, opened 2016-06-13
Don't treat null as zero in chart labels

jbothma:


longhotsummer:
Shouldn't this also apply to the other number formats?

jbothma:
yup. was scared of it but saw a few places needing it so added it now

PR #202, opened 2016-06-13
Fix repmaint calc

jbothma:


longhotsummer:
The amountType field also needs to change. I'll push it.

jbothma:
Fail. Thx

PR #222, opened 2016-06-20
Fix noisy lookups for none instead of None

jbothma:


longhotsummer:
Duh. I misread this from the django docs. Did think it was a bit odd.

PR #224, opened 2016-06-21
Calculate quarterly current ratio for current and previous year, hopi…

jbothma:
…ng that

we'll end up with four quarters.

Uses the latest available month in the quarter. If no month is available in the
quarter, the quarter will be skipped. Still have to add null spacer quarters
for missing quarters and perhaps return only the latest 4 in an ordered array.

### DIFF
--- a/scorecard/profile_data.py
+++ b/scorecard/profile_data.py
@@ -307,7 +307,10 @@ class IndicatorCalculator(object):
                                        'result': None,
                                        'rating': 'bad',
             }) for k in keys][:5]
-        return {'values': values}
+        return {
+            'values': values,
+            'ref': self.references['circular71'],
+        }
 
     def expenditure_trends(self):
         values = {
--- a/scorecard/profile_data.py
+++ b/scorecard/profile_data.py
@@ -50,7 +50,7 @@ class MuniApiClient(object):
             url = self.API_URL + query['cube'] + '/model'
             params = {}
 
-        logger.info("API query %s?%s" % (url, urllib.urlencode(params)))
+        logger.debug("API query %s?%s" % (url, urllib.urlencode(params)))
         return self.session.get(url, params=params, verify=False)
 
     def format_cut_param(self, cuts):
@@ -264,8 +264,7 @@ class IndicatorCalculator(object):
         quarters = {}
         latest_quarter = None
         for year, yeargroup in groupby(year_sorted, year_key):
-            month_sorted =  sorted(yeargroup, key=month_key)
-            month_sorted.reverse()
+            month_sorted = sorted(yeargroup, key=month_key, reverse=True)
             # Loop over months that exist and add their values to quarters
             for month, monthgroup in groupby(month_sorted, key=month_key):
                 monthitems = list(monthgroup)
@@ -277,6 +276,7 @@ class IndicatorCalculator(object):
                     liabilities = [m['amount.sum'] for m in monthitems if m['item.code'] == '1600'][0]
 
                     if quarter_key not in quarters:
+                        result = ratio(assets, liabilities)
                         q = {
                             'date': "%sq%s" % quarter_key,
                             'year': year,
@@ -285,7 +285,8 @@ class IndicatorCalculator(object):
                             'quarter': quarter_idx,
                             'assets': assets,
                             'liabilities': liabilities,
-                            'result': ratio(assets, liabilities),
+                            'result': result,
+                            'rating': 'good' if result >= 1 else 'bad',
                         }
                         quarters[quarter_key] = q
                         if latest_quarter is None:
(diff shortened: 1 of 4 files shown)


scorecard/profile_data.py
```diff
@@ -46,6 +50,7 @@ def api_get(self, query):
             url = self.API_URL + query['cube'] + '/model'
             params = {}
 
+        logger.info("API query %s?%s" % (url, urllib.urlencode(params)))
```
jbothma: should be debug

scorecard/profile_data.py
```diff
@@ -249,6 +254,44 @@ def revenue_breakdown(self):
...
+        year_sorted.reverse()
+        quarters = {}
+        for year, yeargroup in groupby(year_sorted, year_key):
+            month_sorted =  sorted(yeargroup, key=month_key)
+            month_sorted.reverse()
+            # Loop over months that exist and add their values to quarters
+            for month, monthgroup in groupby(month_sorted, key=month_key):
+                monthitems = list(monthgroup)
+                quarter_idx = ((month - 1) / 3) + 1
+                quarter_key = "%sq%s" % (year, quarter_idx)
```
jbothma: could also use a tuple for this if we don't want to use this string in the frontend

scorecard/profile_data.py
```diff
@@ -481,7 +524,24 @@ def noop_structure(self, query, response):
         return response
 
     def get_queries(self):
+        today = datetime.now()
+        in_year_years = [today.year, today.year-1, today.year-2]
```
jbothma: this is quite large right now for development. should reduce

scorecard/profile_data.py
```diff
@@ -249,6 +254,44 @@ def revenue_breakdown(self):
...
+                            'amount_type': 'ACT',
+                            'quarter': quarter_idx,
+                            'assets': assets,
+                            'liabilities': liabilities,
+                            'current_ratio': ratio(assets, liabilities),
+                        }
+                        quarters[quarter_key] = q
+                except IndexError:
+                    print
+                    print("%s %s" % (year, month))
```
jbothma: remove debug prints

scorecard/profile_data.py
```diff
@@ -236,19 +241,79 @@ def revenue_breakdown(self):
...
+        values = []
+        results = self.results['in_year_bsheet']
+        year_key = lambda r: r['financial_year_end.year']
+        month_key = lambda r: r['financial_period.period']
+        year_sorted = sorted(results, key=year_key)
+        year_sorted.reverse()
+        quarters = {}
+        latest_quarter = None
+        for year, yeargroup in groupby(year_sorted, year_key):
+            month_sorted =  sorted(yeargroup, key=month_key)
```
longhotsummer: or `reverse=true` saves you a line

scorecard/profile_data.py
```diff
@@ -236,19 +241,79 @@ def revenue_breakdown(self):
...
+            for month, monthgroup in groupby(month_sorted, key=month_key):
+                monthitems = list(monthgroup)
+                quarter_idx = ((month - 1) / 3) + 1
+                quarter_key = (year, quarter_idx)
+                try:
+                    # Rely on index out of range for missing values to skip month if one's missing
+                    assets = [m['amount.sum'] for m in monthitems if m['item.code'] == '2150'][0]
+                    liabilities = [m['amount.sum'] for m in monthitems if m['item.code'] == '1600'][0]
+
+                    if quarter_key not in quarters:
```
longhotsummer: So we land up using the latest month for which we have data in the current quarter?

scorecard/profile_data.py
```diff
@@ -236,19 +241,79 @@ def revenue_breakdown(self):
...
+                        }
+                        quarters[quarter_key] = q
+                        if latest_quarter is None:
+                            latest_quarter = q
+                except IndexError:
+                    pass
+        # Enumerate the quarter keys we can expect to exist based on the latest
+        keys = []
+        result_quarters = []
+        if latest_quarter is not None:
```
longhotsummer: How about this:

```
year = latest_quarter['year']
q = latest_quarter['quarter']

# generate keys for the quarters we expect
# eg. [(2015, 3), (2015, 2), (2015, 1), (2014, 4), (2014, 3)]
keys = ((year - (1 if x < 0 else 0), x % 4 + 1) for x in xrange(q - 1, q - 6, -1))

# pull matching values, with None entries if missing
values = [quarters.get(k, {default stuff}) for k in keys]
```

scorecard/profile_data.py
```diff
@@ -236,19 +241,79 @@ def revenue_breakdown(self):
...
+    def current_ratio(self):
+        values = []
+        results = self.results['in_year_bsheet']
+        year_key = lambda r: r['financial_year_end.year']
+        month_key = lambda r: r['financial_period.period']
+        year_sorted = sorted(results, key=year_key)
+        year_sorted.reverse()
+        quarters = {}
+        latest_quarter = None
+        for year, yeargroup in groupby(year_sorted, year_key):
```
longhotsummer: I think you could do these two nested loops in one go by making the sort and group key compound: `(year, month)`. Python sorts tuples component-wise, which means `(2015, 4) < (2015, 3) < (2014, 4)`

scorecard/profile_data.py
jbothma: that's right. The month is included in the data passed to the frontend in case we want t say "and in this month we actually used month 8 and not month 9

scorecard/profile_data.py
jbothma: I like how short it is, however for generating the keys, I'm hesitant because it takes me a long time to parse what's actually happening there. I'd rather spend 5 lines for my version that's closer to braindead simple than one that's very easy to break and very hard to debug.

I like your buildup of values array though - so my key generation plus your `values` generation with an array slice (since I generate too many keys) should be a good compromise?

scorecard/profile_data.py
jbothma: fixed

scorecard/profile_data.py
jbothma: fixed

scorecard/profile_data.py
jbothma: fixed

PR #227, opened 2016-06-21
Contractor and staff wages no rating if None

longhotsummer:


longhotsummer:
Turns out that in python `None < 10` (or any number)

PR #230, opened 2016-06-21
Link from expenditure chart to source

longhotsummer:


jbothma:
so the reason I didn't link it when I linked income is because we didn't have the table. but like current ratio, a broken link still has more information than nothing, so let's go with this and add functional breakdown to table view ASAP

longhotsummer:
You mean we don't have the functional breakdown? Agreed. Let's put what we have and then improve.

PR #253, opened 2016-06-27
standard operating procedure

jbothma:


jbothma:
@longhotsummer I think this is at a good place for review now. At least the README

longhotsummer:
Taking a look.

README.md
```diff
@@ -42,6 +42,61 @@ Data import is still a fairly manual process leveraging the DB and a few SQL scr
 
 *Remember to run `VACUUM ANALYSE` or REINDEX tables after significant changes to ensure stats are up to date to use indices properly.*
 
+# Standard Operating Procedure
+
+This covers how to keep the data up to date. Each quarter, as new data is released, the following needs to be done to update the data served by the API and the Citizen Scorecard. It's best to do this on a test database first and verify the results before updating the production database.
+
+## Extract CSV datasets from Excel Spreadsheets
```
longhotsummer: I'd put the CSV stuff first. That's the bulk of the data.

longhotsummer:
This is great.

Have you been able to test the CSV?

Do the production tables have the new constraints (NOT NULL, etc.)?

README.md
jbothma: These scripts are necessary to produce the CSVs for the non-financials. They should be updated using similar scripts but I didn't write them now because I don't have new data for them.

jbothma:
yeah - the SQL was added as I executed each locally and then against prod.

I didn't add NOT NULL constrains all over the place in the end. I forgot about it after I got the initial couple of scripts working. I fix.

PR #255, opened 2016-06-29
Make zero vs null clearer in table view

longhotsummer:
- == null
0 == 0

I'm not sure if this will be clear. Thoughts?


longhotsummer:
I still want to put a comment into the footer about this. Let's keep it open.

jbothma:
what does centre dot mean in mathematics? also zero? or blank? late feedback. But I think we could use a firetruck for null as long as we define it somewhere. I think the distinction is good. Also a link when we have the FAQ up would be good.

longhotsummer:
I think · would be a great indicator for null, better that "-" which is sometimes used as zero.

PR #257, opened 2016-06-30
Replace example screenshots on homepage

fritzjooste:
Replace example screenshots with cash balance indicator screenshots


longhotsummer:
great.

PR #260, opened 2016-07-04
User can choose function in table view

longhotsummer:


longhotsummer:
This doesn't support function in downloads yet

PR #261, opened 2016-07-12
Group table view by function

longhotsummer:
* also ensure you keep the chosen munis when the dataset is changed
* downloads include functions if we're filtering by them

### DIFF
--- a/municipal_finance/static/javascript/table.js
+++ b/municipal_finance/static/javascript/table.js
@@ -658,10 +658,17 @@
       var toHighlight = [];
       var self = this;
 
-      // group by code then municipality
+      // group cells by item code then municipality
       cells = _.groupBy(cells, 'item.code');
       _.each(cells, function(items, code) {
-        cells[code] = _.indexBy(items, 'demarcation.code');
+        cells[code] = _.groupBy(items, 'demarcation.code');
+
+        // group by function?
+        if (functions.length > 0) {
+          _.each(cells[code], function(items, muni) {
+            cells[code][muni] = _.indexBy(items, 'function.code');
+          });
+        }
       });
 
       // values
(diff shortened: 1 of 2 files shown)


longhotsummer:
@jbothma when you have a sec?

municipal_finance/static/javascript/table.js
```diff
@@ -539,10 +551,10 @@
         // copy this, we're going to change it
         params.drilldown = params.drilldown.slice();
 
-        // ensure the download has all attributes, except function
-        // which we'll deal with later
+        // ensure the download has all relevant attributes.
+        // we only include functions if we're already filtering on functions
         _.each(cube.model.dimensions, function(dim, dim_name) {
-          if (dim_name != 'function') {
+          if (dim_name != 'function' || hasFunctions) {
```
jbothma: generally I try to refer to govt. functions when space allows else I end up thinking we're dealing with higher order functions for a while, but that can become verbose

municipal_finance/static/javascript/table.js
```diff
@@ -658,10 +670,17 @@
...
+      // group cells by item code then municipality
       cells = _.groupBy(cells, 'item.code');
       _.each(cells, function(items, code) {
-        cells[code] = _.indexBy(items, 'demarcation.code');
+        cells[code] = _.groupBy(items, 'demarcation.code');
+
+        // group by function?
+        if (functions.length > 0) {
+          _.each(cells[code], function(items, muni) {
+            cells[code][muni] = _.indexBy(items, 'function.code');
```
jbothma: I obviously didn't use govt_function in the API - fail

jbothma:
LGTM except adding Government to the comment as above. Would you like me to try it? But I'm happy having a rough idea how the code works now

longhotsummer:
I've added some extra comments, the functions does get confusing.

PR #281, opened 2016-07-31
Change way that video modals are loaded

fritzjooste:


longhotsummer:
This breaks the language chooser for the about-muni-finance video. Instead, I've borrowed from this idea and added a second popup for single-language videos. See #282.

PR #284, opened 2016-08-15
Liquidity ratio

jbothma:
Seems to work alright. This is of course super clearly defined... By which I mean I don't know what to cite yet.
I can't match this well to everything yet but Kouga and NMA seemed to match fairly well


longhotsummer:
LGTM

PR #298, opened 2016-09-13
Link from "Not available" in scorecard to FAQ entry

jbothma:
I don't think it's clear where to find the explanation of the asterisk - especially since it's not clear it's hyperlinked.

not-avail-astrerisk

But the underline on the subscript doesn't look good

not-avail-asterisk-hover

The text looks kinda useful, but without underline it looks more like a heading for the description below

not-avail-text

I think with underline (always, not just on hover) it works - even though it makes the indicator busier.

not-avail-text-hover

### DIFF
--- a/scorecard/static/stylesheets/scorecard.scss
+++ b/scorecard/static/stylesheets/scorecard.scss
@@ -207,7 +207,7 @@ blockquote {
       margin-left: 30px;
     }
   }
-  
+
   .search {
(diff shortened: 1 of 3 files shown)


longhotsummer:
"What does this mean" is my choice, much clearer than the asterisk.

I think it shouldn't be underlined, though, since that's the idiom throughout the page.

PR #308, opened 2016-10-24
Hide out of date political appointments and explain

jbothma:
Don't show mayor or deputy mayor since those positions are out of date since the local elections. Keep the places to indicate that this information exists and will come soon.

Explain a) that it's updated quarterly and b) that they're hidden because of the recent elections.

I think mentioning "political appointments" works a little toward educating that mayor and deputy are appointed differently from managers.

muni-money-mayoral-contacts-2016


longhotsummer:
LGTM

PR #323, opened 2016-11-02
Support the audit results cube

longhotsummer:


longhotsummer:
The items show up on one line, which isn't ideal but it works okay for now. An option would be to just show all years.

jbothma:
:+1: 

I think showing years as a row heading would add a lot of complexity to the code. I think getting other cubes working is a higher priority than showing more years of this cube. Let's see what things look like after adding those.

longhotsummer:
Agreed.

PR #330, opened 2016-11-17
Profile page tweaks

longhotsummer:
Adjustments to positioning and size to make the page easier to read. Gathered while working on mockups for comparisons.

* smaller plus alongside 'show calculation'
* move chart down a bit so that indicator value stands out
* smaller key (key and 'did you know' text is now the same size), so that key is secondary and not primary
* key shifted left slightly so that text aligns (less visual noise from different vertical lines)
* less **heavy bold** text

Before:
cpt-before

After
cpt-after

### DIFF
--- a/scorecard/static/stylesheets/scorecard.scss
+++ b/scorecard/static/stylesheets/scorecard.scss
@@ -1312,6 +1312,7 @@ blockquote {
 // Charts
 
 .chart-container {
+  margin-top: 40px;
   font-size: 12px;
 
   rect.chart-bar,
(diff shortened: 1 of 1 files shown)


longhotsummer:
@lailahr @jbothma curious what you think?

jbothma:
great changes.

have you thought of aligning the key to right? At this size it would associate the key better with the chart. On narrow devices it wouldn't look different. When the description column is quite wide it might look weird.

longhotsummer:
I'm going to push this and then explore moving the key as part of the descriptive comparison changes.

PR #345, opened 2016-12-03
Use median of absolute values of over- or underspending

jbothma:
I'm not sure about using the median of absolutes of over/underspending.

If we phrase it as "the [median] difference between budget and actual spending" it makes sense but when phrased just as over or underspending the picture in my head is that this median (and its sign) is on the same spectrum as the numbers shown here. For example, the median for raw over/under was negative while cape town was positive in 2014. now we show a positive number in the tooltips because of abs() in each case which is also a bit weird.

I'm not sure what's correct and what's clear to the user.

munimoney-overunder-abs-median


scorecard/profile_data.py
```diff
@@ -639,6 +639,21 @@ def get_indicators(api_data):
     return indicators
 
 
+def abs_median(items):
+    return median([abs(x) for x in items])
```
longhotsummer: Hmm. I think what we should be doing here is finding the index of the median element (absolute values), then return that actual element (without the median).

A bit like a `decorate / sort / undecorate` pattern, but applied to calculating the median. My reasoning is that the negative value informs whether it's under or overspending, it doesn't actually impact the *degree* of over/underspending (which is what we're trying to get when we calculate the median).

jbothma:
outcomes of whiteboarding:
- showing the sign of a median that was selected from absolutes of results doesn't mean anything - the sign becomes accidental
- we've combined overspending and underspending into one chart - a muni will either overspend or underspend in a given period - now it gets interesting when there coud be some that overspend and some that underspend
- what's currently in prodution isn't wrong - it's a particular view of the data. It tells the user whether similar munis typically overspend or underspend, but it doesn't say anything about the degree at which either happens
- another view of the distribution is to take the median of the absolutes, as in this PR - it doesn't give an indication of whether over- or underspending typically happens, it tells the user to what degree munis tend to stray from their budget in whatever direction
- a third view would be to take two medians - an overspending median and an underspending median. We propose showing the median relevant to whatever happened in that period - if a muni underspent we show the underspending median. If no one else underspent in that period, we point that out, and perhaps show the overspending median.

PR #347, opened 2016-12-07
Disestablished notice

longhotsummer:
Notice on page for disestablished munis.


longhotsummer:
This breaks the page if our geography table doesn't have the new munis.

PR #351, opened 2016-12-07
Established notice

longhotsummer:
Notices for newly established munis that are missing indicators.


longhotsummer:
Demarcation element in indicator dict updated to support `established`:

```json
    "demarcation": {
        "established": true,
        "established_date": "2099-01-01",
        "established_from": ["BUF", "TSH"]
    },
```

longhotsummer:
city of cape town - municipal money 2016-12-07 17-16-19

PR #373, opened 2017-06-26
Update dev env

jbothma:
- ubuntu 14.04 wasn't building any more
- ubuntu 16.04 comes with modified debian python with ssl3 flag dropped
- updated genevent to use python > 2.7.8
- added docker-compose file to make dev env management more consistent
- wazimap now needs settings.TESTING to be defined
- wazimap-mapit required GDAL so this is still sometimes the easiest way to get a dev env


Dockerfile
```diff
@@ -13,7 +13,8 @@
 
 
 
-FROM ubuntu:trusty
+#FROM ubuntu:trusty
+FROM gdal-testy
```
jbothma: remove hack

PR #390, opened 2018-01-10
Upgrade to django 1.11

longhotsummer:


municipal_finance/wsgi.py
```diff
@@ -16,6 +16,6 @@
 patch_psycopg()
 
 from django.core.wsgi import get_wsgi_application
-from whitenoise.django import DjangoWhiteNoise
+# from whitenoise.django import DjangoWhiteNoise
```
longhotsummer: if we don't need these lines, let's just remove 'em. ✂️

requirements.txt
```diff
@@ -1,17 +1,17 @@
 appnope==0.1.0
-git+git://github.com/Code4SA/babbage
+babbage==0.1.1
```
longhotsummer: @jbothma pls sanity check?

requirements.txt
```diff
@@ -32,12 +32,14 @@ pathlib==1.0.1
 pathlib2==2.1.0
 pexpect==4.0.1
 pickleshare==0.7.2
+pkg-resources==0.0.0
```
jbothma: is this a typo? This broke `pip -U -r requirements.txt` for me

requirements.txt
jbothma: This won't work. The pip babbage 0.1.1 (https://github.com/spendb/babbage.git) is before our star schema changes. That got merged and tagged as babbage 0.2.1 but that doesn't work out of the box either.
I suggest for this PR to stick with out babbage from github. babbage 0.2.1 did bring python 3 support and it'd be nice to switch to the upstream babbage if we can.

A nice way to test that the API works as expected is to follow the profile data update instructions, i.e. running `python bin/materialised_views.py --api-url http://localhost:8000/api  --profiles-from-api` and doing `git diff` to see if the profiles look like they did before. If something changed from what's in git, something's probably broken.

Also watch out - looks like they brought in their own csv generation in 2.1 so we should see if that'll conflict with ours or if that'll make ours redundant. I think minor changes in presentation between our csv response and theirs is acceptable if it lets us remove code and if the CSV download links still work

longhotsummer:
ship it!

PR #392, opened 2018-01-17
Upgrade python

ebsuku:
I've made the necessary changes to get py3 working.


longhotsummer:
Awesome!

requirements.txt
```diff
@@ -9,13 +9,11 @@ django-extensions==1.6.1
...
 django-wkhtmltopdf==3.0.0
 ecdsa==0.13
-enum34==1.0.4
 Flask==0.10.1
-functools32==3.2.3.post2
 futures==3.0.5
 gevent==1.2.1
 gnureadline==6.3.3
-grako==3.8.1
+grako==3.10.1
```
jbothma: Please remove pkg-resources=0.0.0 - apparently this is a bug in ubuntu
This breaks installing deps from requirements.txt, at least on non-ubuntu https://stackoverflow.com/questions/39577984/what-is-pkg-resources-0-0-0-in-output-of-pip-freeze-command

jbothma:
I think we need to make sure the default page size is what it used to be, or at the very least make the scorecard specify page size

I tested this by setting export API_URL=http://localhost:8002/api for SITE_ID=3

Expected:
munimoney-pagesize-prod

This branch:
munimoney-python3-pagesize

jbothma:
CSV download seems to be broken
http://localhost:8002/api/cubes/cflow/aggregate?drilldown=amount_type.label%7Camount_type.code%7Cfinancial_period.period%7Cperiod_length.length%7Cfinancial_year_end.year%7Citem.composition%7Citem.return_form_structure%7Citem.position_in_return_form%7Citem.label%7Citem.code%7Cdemarcation.label%7Cdemarcation.code%7Cdemarcation.code%7Cdemarcation.label%7Citem.code%7Citem.label%7Citem.return_form_structure&cut=financial_year_end.year%3A2016%7Camount_type.code%3AAUDA%7Cfinancial_period.period%3A2016%7Cdemarcation.code%3A%22DC44%22&aggregates=amount.sum&page=1&pagesize=50&order=demarcation.code%3Aasc%2Citem.position_in_return_form%3Aasc&format=csv
munimoney-csv-download

jbothma:
It concerns me a bit that they dropped `type` from the model in their tests 
```
--- a/tests/fixtures/models/cra.json
+++ b/tests/fixtures/models/cra.json
@@ -49,32 +49,29 @@
         },
         "cofog2": {
             "attributes": {
                 "change_date": {
                     "column": "cofog2_change_date",
                     "label": "Change date",
                     "type": "date"
                 },
                 "description": {
                     "column": "cofog2_description",
-                    "label": "Description",
-                    "type": "string"
+                    "label": "Description"
                 },
                 "label": {
                     "column": "cofog2_label",
-                    "label": "Label",
-                    "type": "string"
+                    "label": "Label"
                 },
                 "name": {
                     "column": "cofog2_name",
-                    "label": "Name",
-                    "type": "integer"
+                    "label": "Name"
```

I can't remember off the top of my head what this was for - if it was just intended to help when types weren't inferred properly, or if they totally dropped support for that. Which would at best mean our models now have a superfluous field and at worst that the API might do something weird with types.

jbothma:
```
+++ b/babbage/query/pagination.py
@@ -1,18 +1,20 @@
 from babbage.query.parser import Parser
 from babbage.util import parse_int
 
 
 class Pagination(Parser):
     """ Handle pagination of results. Not actually using a parser. """
 
-    def apply(self, q, page, page_size, page_max=10000):
+    def apply(self, q, page, page_size, page_max=None, page_default=100):
         page_size = parse_int(page_size)
         if page_size is None:
-            page_size = page_max
+            page_size = page_default
+        if page_max is None:
+            page_max = 10000
```

ebsuku:
Alright.
Downloading of csv corrected.
pkg-resources removed

scorecard/profile_data.py
```diff
@@ -1046,10 +1048,10 @@ def get_muni_specifics(cls, api_data):
         # If latest is missing, there are none to show.
         if latest_quarter is not None:
             keys = []
-            for q in xrange(latest_quarter['quarter'], 0, -1):
-                keys.append((latest_quarter['year'], q))
-            for q in xrange(4, 0, -1):
-                keys.append((latest_quarter['year']-1, q))
+            for q in range(int(latest_quarter['quarter']), 0, -1):
```
jbothma: I think we should rather fix quarter_tuple(...) to return ints as one would expect, rather than turn all of the floats it's generating into ints individually

scorecard/profile_data.py
```diff
@@ -1198,16 +1200,16 @@ def get_muni_specifics(cls, api_data):
...
-            except KeyError, e:
+            except KeyError as e:
                 logger.debug(e)
 
         # Enumerate the quarter keys we can expect to exist based on the latest
         # If latest is missing, there are none to show.
         if latest_quarter is not None:
             keys = []
-            for q in xrange(latest_quarter['quarter'], 0, -1):
+            for q in range(int(latest_quarter['quarter']), 0, -1):
```
jbothma: ditto

scorecard/profile_data.py
```diff
@@ -1492,7 +1494,7 @@ def calendar_to_financial(year, month):
 
 
 def quarter_idx(month):
-    return ((month - 1) / 3) + 1
+    return ((month - 1) // 3) + 1
```
jbothma: I'm not sure floor is what we wanted here, but maybe! In my head this was Python2 5/3 = 2 or Python3 round(5/3) = 2 but apparently it was Python2 5/3 = 1 so perhaps this is fine...

http://plasmasturm.org/log/6debug/

ebsuku:
I remember first fixing  the range issue by putting int() there, then deciding later to fix it in quarter_idx as the code would be a bit cleaner.

scorecard/profile_data.py
longhotsummer: This looks fine to me. It calculates the quarter from a 1-based month count. the `//` does integer division.

PR #440, opened 2020-07-10
non-financial Quarterly Update for 2020-07

juriejan:


scorecard/materialised/profiles/DC20.json
```diff
@@ -972,10 +972,15 @@
                 "url": "http://mfma.treasury.gov.za/Circulars/Pages/Circular71.aspx"
             },
             "values": [
+                {
+                    "date": 2019,
+                    "rating": null,
+                    "result": null
+                },
```
jbothma, line 979: Looks like UIFW is always null for 2019 - was there 2019 data for these munis? did we load it?

PR #448, opened 2020-09-30
Admin interface for updating contacts

juriejan:


docker-compose.dev.yml
```diff
@@ -0,0 +1,101 @@
+version: '3'
```
jbothma, line 1: why a dev dockerfile when the main dockerfiles are intended for dev already?

municipal_finance/models.py
```diff
@@ -326,7 +335,7 @@ class Meta:
 
 
 class MunicipalityStaffContacts(models.Model):
-    demarcation_code = models.TextField()
+    demarcation_code = models.TextField(primary_key=True)
```
jbothma, line 338: watch out - I suspect this is a composite unique key between demarcation_code and role - looking at the comparison on import https://github.com/OpenUpSA/municipal-data/blob/master/sql/upsert/municipality_staff_contacts.sql#L24

and 
```
municipal_finance=> select demarcation_code, role from municipality_staff_contacts order by demarcation_code limit 5;
 demarcation_code |             role             
------------------+------------------------------
 BUF              | Deputy Mayor/Executive Mayor
 BUF              | Municipal Manager
 BUF              | Speaker
 BUF              | Mayor/Executive Mayor
 BUF              | Chief Financial Officer
```

municipal_finance/models.py
```diff
@@ -383,12 +402,31 @@ class UifwexpFacts(models.Model):
...
+class MunicipalityProfilesRebuild(models.Model):
+    user = models.ForeignKey(User)
+    datetime = models.DateTimeField(auto_now_add=True)
+
+    class Meta:
+        db_table = 'municipality_profiles_rebuild'
+
+
+class MunicipalityProfile(models.Model):
+    demarcation_code = models.CharField(max_length=10, primary_key=True)
```
jbothma, line 428: could also do a foreign key on geography but not a big deal

requirements.txt
```diff
@@ -12,6 +12,7 @@ django-cors-headers==2.1.0
 django-elasticsearch-dsl==7.1.0
 django-elasticsearch-dsl-drf==0.20.5
 django-extensions==1.6.1
+django-environ==0.4.5
```
jbothma, line 15: yes! the future!

docker-compose.dev.yml
juriejan: Just used an extra file to play around with for a bit. Merged them now.

municipal_finance/models.py
juriejan: Yes. There is a `unique_together` defined in the `Meta` class for the model. I had to add a single primary key to be use the model for querying and creating. Sure you know this but, usually Django models would have an `id` primary key and then unique constraints in these cases. I see in the `sql` folder most of the tables have a `id serial` column defined, but I don't see it in the database. What's up with that? Are those scripts relevant? Can we being adding `id` columns for the models we need to work with?

municipal_finance/models.py
juriejan: I like that idea. Should we perhaps give this model an `id` for the primary key and change to a foreign key, or use a `OneToOneField`?

jbothma:
Yeah - the models were created after the tables and I didn't bring them up
to django spec. Please add id columns as needed.

jbothma:
Whatever works best for you is good for me!

municipal_finance/management/commands/sample.py
```diff
@@ -0,0 +1,52 @@
+from django.core.management.base import BaseCommand, CommandError
+
+from municipal_finance.models import UifwexpFacts, AuditOpinionFacts
+
+class Command(BaseCommand):
+    help = 'Sample command'
```
jbothma, line 6: do we still need this?

municipal_finance/models.py
jbothma: How can this allow the following data:

```csv
demarc,role,name
BUF,Boss,John
BUF,Prole,Peter
```
- doesn't primary key imply unique?

jbothma:
cool - remember to get the test running in travis!

Let's get it into staging!

municipal_finance/models.py
juriejan: I've gone with added an `id` column as mentioned in the previous message.

PR #458, opened 2020-11-06
Feature - Initial MSCOA indicators

juriejan:


models/bsheet_mscoa.json
```diff
@@ -0,0 +1,136 @@
+{
+    "fact_table": "bsheet_facts_mscoa",
```
jbothma, line 2: what do you think of naming these something like `bsheet_v2` instead of `bsheet_mscoa` and including something like `Section 71-level aggregation of mSCOA data from 2019-20 onwards replacing the original bsheet cube`

jbothma: Im pondering this since this data isn't actually clasified according to the mSCOA schema, it's classified according to Section 71 version something, based on aggregated mSCOA data. Calling it mSCOA might be confusing and disappointing to experts

municipal_finance/compile_data.py
```diff
@@ -270,6 +277,22 @@ def compile_rating_counts(api_url):
...
 def compile_data(api_url):
-    compile_profiles(api_url)
-    compile_medians(api_url)
-    compile_rating_counts(api_url)
+    # Setup the client
+    http_client = FuturesSession(
+        executor=ThreadPoolExecutor(max_workers=10)
+    )
+    http_client.mount(
+        "http://",
```
jbothma, line 285: not https?

municipal_finance/settings.py
```diff
@@ -362,7 +361,8 @@
     "queue_limit": 100,
     "bulk": 50,
     "orm": "default",
-    "poll": 10,  # Check for queued tasks this frequently (seconds)
+    "poll": 10,
```
jbothma, line 364: why drop the doc?

models/bsheet_mscoa.json
juriejan: Makes sense. Making those changes now.

municipal_finance/settings.py
juriejan: I thought that if you'd like to understand the configuration of DjangoQ you can go to the documentation. I usually like to keep comments for information specific to what is being implemented, and not use it for documentation that has a place elsewhere.

models/bsheet_mscoa.json
juriejan: Change made @jbothma. Please resolve if you're happy.

municipal_finance/compile_data.py
juriejan: Copy and pasted from old code. Changed.

municipal_finance/settings.py
```diff
@@ -340,7 +339,7 @@
     "root": {"handlers": ["console"], "level": "ERROR"},
     "loggers": {
         "municipal_finance": {"level": "DEBUG" if DEBUG else "INFO"},
-        "sqlalchemy.engine": {"level": "INFO" if DEBUG else "WARN"},
+        # "sqlalchemy.engine": {"level": "INFO" if DEBUG else "WARN"},
```
jbothma, line 342: is this intentional or forgotten temporary comment?

municipal_finance/settings.py
jbothma: fair point - I tend to document things like this, or at least the unit, so you don't have to look that up, but I don't feel very strongly about it

municipal_finance/compile_data.py
jbothma: if you want this to work both in dev and prod/staging you might need an _http_ and an _https_ mount :)

But perhaps we could just pop api_url un there if it works as a prefix?

models/cflow_v2.json
```diff
@@ -0,0 +1,136 @@
+{
+    "fact_table": "cflow_facts_v2",
+    "label": "Cash Flow (V2)",
```
jbothma, line 3: Do these new cubes show up in https://municipaldata.treasury.gov.za/ automatically? Do they work? Should we exclude them somehow so we don't show links to broken things until we can make sure they work properly?

municipal_finance/settings.py
juriejan: Guess I forgot it there. It speeds up a test run during development not to output all the queries.

municipal_finance/models/income_expenditure.py
```diff
@@ -0,0 +1,108 @@
...
+        models.DO_NOTHING,
+    )
+
+    class Meta:
+        db_table = 'incexp_facts_v2'
+        indexes = [
+            models.Index(fields=['item_id']),
+            models.Index(fields=['amount_type_id']),
+            models.Index(fields=['function_id']),
+        ]
```
jbothma, line 91: postgres or django automatically creates indices on foreign key fields - this might result in duplicate index maintenance.

scorecard/fixtures/indicators/bsheet_facts_v2.csv
```diff
@@ -0,0 +1,97 @@
+id,demarcation_code,period_code,amount,financial_year,period_length,financial_period,amount_type,item
```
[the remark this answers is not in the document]
juriejan: These are fixtures for the indicator tests. Perhaps I should include 'test' somewhere in the names?

municipal_finance/resources.py
```diff
@@ -0,0 +1,60 @@
...
+class BsheetFactsV2Resource(resources.ModelResource):
+    class Meta:
+        model = BsheetFactsV2
+        import_id_fields = ['id']
+
+
+class BsheetFactsV1Resource(resources.ModelResource):
+    class Meta:
+        model = BsheetFactsV1
+        import_id_fields = ['id']
```
jbothma, line 60: is this just for dev? we wouldn't have data with ids in normal operations - this makes me a bit nervous that it's a dev-specific thing but not clearly marked as such, and in normal production code.

jbothma (commented):
mostly questions - you've done a ton of work!!

I will have to continue later today

municipal_finance/compile_data.py
juriejan: It seems to be working fine on dev and staging currently, but will definitely fix if it stops working.

models/cflow_v2.json
juriejan: They do show up. I've added a commit that excludes them from being listed on the data portal.

municipal_finance/compile_data.py
juriejan: Apologies. It wasn't working for dev. Commit with fix incoming.

municipal_finance/resources.py
juriejan: These are only used in tests, and will probably not be used for uploading real data (too slow and memory hungry). I have however made a commit that indicated the correct fields to lookup on.

municipal_finance/models/income_expenditure.py
juriejan: Commit made that removes the keys.

municipal_finance/settings.py
jbothma: might be good to use a different variable than DEBUG so it's easier to tune log level, but not urgent right now

scorecard/fixtures/indicators/bsheet_facts_v2.csv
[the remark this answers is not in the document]
jbothma: Yeah I think it's really important to have the word "test" in the name or path. No strong feelings on where it's best-placed. I usually try to name the test file the same as the test module, class or method and locate it nearby (except when I use django-style dumpdata/loaddata/`TestCase.fixtures` style fixtures )

municipal_finance/resources.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class BsheetFactsV1Resource(resources.ModelResource):
+    class Meta:
+        model = BsheetFactsV1
+        import_id_fields = [
+            'demarcation_code',
+            'period_code',
+            'item_code',
+        ]
```
jbothma, line 68: This is not something that should accidentally be used in production, because import_id_fields does not guarantee that the whole "import set" is deleted - so a mistaken row in an old upload won't be removed when that period is re-uploaded with a correction.

I don't think something that might accidentally be used incorrectly should be in prod code - perhaps we can move it to test code where we can rely on the test framework to flush the DB between tests.

jbothma:
Other than the comments above about keeping test data and code separate from production stuff, this looks good to go! I'll assume you've checked that the quarterly indicator values affected by this change match production.

scorecard/fixtures/indicators/bsheet_facts_v2.csv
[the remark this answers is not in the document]
juriejan: Moved them into a 'tests' directory. The naming begins to get closer to what you mention here in later PRs.

municipal_finance/resources.py
juriejan: As far as I understand what you're saying, and how resources import works, the 'mistaken' rows will indeed be replaced, as long as the demarcation_code, period_code and item_code are the same.

Also keep in mind that these can only be used to import data that was directly exported from our database, and not datasets provided by National Treasury.

These classes also don't show anything on admin yet. Just in case you though that might be the case.

I get what you're saying about keeping things that are only used by tests with tests, and I'll move more in that direction going forward.

Can we leave it here for now?

juriejan:
@jbothma I'll do a final check to see that nothing is affected by the update if there is no data available. Do you mind if I squash this branch into one commit?

jbothma:
> I'll do a final check to see that nothing is affected by the update if there is no data available. Do you mind if I squash this branch into one commit?

No problem

PR #459, opened 2020-11-16
Disable preloading cubes for management commands to prevent errors wi…

juriejan:
…th non-existing tables


jbothma (approved):

jbothma:
Nice. what a beast this thing is...

PR #473, opened 2020-12-03
Indicators compatible with mSCOA-based v2 cubes

jbothma:


municipal_finance/migrations/0024_delete_auditopinions.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('municipal_finance', '0023_capital_initial_data'),
+    ]
+
+    operations = [
+        migrations.DeleteModel(
```
jbothma, line 15: why?

municipal_finance/resources.py
```diff
@@ -140,3 +146,54 @@ class Meta:
...
+        import_id_fields = [
+            'demarcation_code',
+            'period_code',
+            'item',
+            'function',
+            'capital_type',
+        ]
+
+
+class UIFWExpenditureFactsResource(resources.ModelResource):
```
jbothma, line 192: why do we have facts resources under production code? This makes me nervous

municipal_finance/resources.py
```diff
@@ -140,3 +146,54 @@ class Meta:
...
+class CapitalFactsV2Resource(resources.ModelResource):
+    class Meta:
+        model = CapitalFactsV2
+        import_id_fields = [
+            'demarcation_code',
+            'period_code',
+            'item',
+            'function',
+            'capital_type',
+        ]
```
jbothma, line 189: why these if the data has an id field?

scorecard/profile_data/indicators/capital_budget_spending.py
```diff
@@ -0,0 +1,99 @@
+from .indicator_calculator import IndicatorCalculator
+from .utils import (
+    group_by_year,
+    populate_periods,
+    filter_for_all_keys,
+    percent,
+)
+
+
+class CapitalBudgetSpending(IndicatorCalculator):
```
jbothma, line 10: meer technically is dit "capital spending budget vs actual' of "capital over/under-spending"

jbothma: spending beteken nie noodwendig actual nie - although daar is een of ander convention van expenditure vs spending 

So eks nie seker hierdie naam is duideliker nie.

jbothma: wat van CapitalOverUnderSpending

ons wil spending in die naam he want dis nie revenue/income nie
ons wil die notion van over/under of budget vs actual (nie heeltemal nie) of difference daar he.
ek dink over/under is die mees duidelike opsie.

municipal_finance/migrations/0024_delete_auditopinions.py
juriejan: Wow! I've got no idea where this came from. Getting rid of it.

juriejan: Ah I forgot @jbothma! It was an unused model. We've got AuditOpinionFacts for the actual data. That good with you?

municipal_finance/resources.py
juriejan: @jbothma do you means we should only use `demarcation_code` and `period_code`? What `import_id_fields` does is identify the record to replace. Pretty much the same way the upsert scripts work. Perhaps knowing this might make you less nervous about having these in production code?

municipal_finance/migrations/0024_delete_auditopinions.py
jbothma: aaah cool - I guess the build would have broken if this table was used but missing anyway

municipal_finance/resources.py
juriejan: @jbothma I've moved the resources for facts to the test folder where they are used.

municipal_finance/resources.py
jbothma: for this to be safe in production, you need to replace everything with the same demarcation_code and period_code ignoring item, function and capital_type - since you can't guarantee that you will get a replacement row with the same item, function and capital type

jbothma: Screenshot_2020-12-05_10-14-35

I mean your test data has an id column. Of course this is fine, just thought I'd bring it to your attention if you intended to use just classifying columns or just unique id

juriejan: Ah! I realised the other day that the items need to have to have consistent id's to ensure the tests always refer to the correct items. I'm going to include id's in the initial fixtures. That would still not make this dependable for production import, but then I've already moved these resources to the tests.

juriejan: @jbothma I'm thinking now that I'm going to stop using resources for data import over the course of the next few days, switching to using the same code that I'll be using to import in the production code (for admin imports). In that case the stability of the tests should be fine without `id`s. Are you fine with keeping it like that for now?

juriejan: @jbothma Ah those ids aren't used at all. Just did a quick export from the DB for these. I'll exclude them from the tests now.

scorecard/profile_data/indicators/capital_budget_spending.py
juriejan: @jbothma my thinking here (which is a rule I try to apply consistently, within reason) is to keep the names that we use on the back-end as close as possible to what we see on the front-end. Would you still like me to adjust then names to resemble more closely to what they actually represent?

jbothma: yeah - I think the top priority is that the code is clear about what it represents - this change is making it less clear.

I agree with backend and frontend names matching, so I would propose we call this `CapitalOverUnderSpend` now - intentionally capitalising S so it's clearer that it relates to both the over and under, and if you would like to make frontend and backend names match, rather get it fixed in the frontend when you and/or matt have time, than make the backend match the unclear frontend name.
ping @juriejan

jbothma: same for operating over/under-spend section

municipal_finance/resources.py
jbothma: yeah - totally happy with using the new import functionality when that's available - nice one

jbothma: @juriejan

scorecard/profile_data/indicators/capital_budget_spending.py
juriejan: Makes sense @jbothma. I'll create a task for reviewing the titles of the indicators.

.vscode/.ropeproject/config.py
```diff
@@ -0,0 +1,114 @@
+# The default ``config.py``
```
jbothma, line 1: can we .gitignore this?

PR #487, opened 2020-12-11
Ability to update cube facts tables via admin GUI (income and expenditure, repairs and maintenance, cash flow)

juriejan:


.vscode/.ropeproject/config.py
```diff
@@ -0,0 +1,114 @@
+# The default ``config.py``
+# flake8: noqa
+
+
+def set_prefs(prefs):
```
jbothma, line 5: would be nice to add .vscode to gitignore

municipal_finance/admin.py
```diff
@@ -62,26 +65,95 @@ def save_model(self, request, obj, form, change):
...
+        if obj is None:
+            return ('user',)
+        else:
+            return super(BaseUpdateAdmin, self).get_exclude(request, obj)
+
+
+@admin.register(IncomeExpenditureV2Update)
+class IncomeExpenditureV2UpdateAdmin(BaseUpdateAdmin):
+
+    def save_model(self, request, obj, form, change):
```
jbothma, line 108: we might want to check if it's new or an edit later using `change` I think, since you probably don't want changes that come about who knows how to accidentally re-import and perhaps overwrite later files.

but that can be added to the backlog to do later.

municipal_finance/models/updates.py
```diff
@@ -0,0 +1,55 @@
...
+    datetime = models.DateTimeField(auto_now_add=True)
+    deleted = models.BigIntegerField(null=True)
+    inserted = models.BigIntegerField(null=True)
+
+    class Meta:
+        abstract = True
+
+
+class MunicipalStaffContactsUpdate(BaseUpdate):
+    file = models.FileField(
```
jbothma, line 19: it might be a good idea to add a timestamp or uuid to the file name to ensure it's unique and that we don't overwrite one file with another and then get confused when we're debugging something.

https://github.com/vulekamali/datamanager/blob/master/budgetportal/models.py#L1864

I'm happy for that to be added to the backlog though and merging this as is

municipal_finance/resources.py
```diff
@@ -149,3 +154,55 @@ class AgedCreditorItemsV2Resource(resources.ModelResource):
     class Meta:
         model = AgedCreditorItemsV2
         import_id_fields = ['code']
+
+
+class IncexpFactsV1Resource(resources.ModelResource):
```
jbothma, line 159: I know I'm a stuck record but I really think facts resources should be kept in a clearly tests-only part of the codebase so that no one accidentally thinks these are ok to use in production

municipal_finance/settings.py
```diff
@@ -395,7 +395,7 @@
     "root": {"handlers": ["console"], "level": "ERROR"},
     "loggers": {
         "municipal_finance": {"level": "DEBUG" if DEBUG else "INFO"},
-        "sqlalchemy.engine": {"level": "INFO" if DEBUG else "WARN"},
+        # "sqlalchemy.engine": {"level": "INFO" if DEBUG else "WARN"},
```
jbothma, line 398: dev comment

municipal_finance/update/cash_flow_v2.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def row_to_obj(amount_types, items, row: CashFlowRow):
+    (
+        financial_year,
+        amount_type_code,
+        period_length,
+        financial_period
+    ) = period_code_details(row.period_code)
+    amount = int(row.amount) if row.amount else None
```
jbothma, line 53: why int? I'm pretty sure floats or decimals should be allowed depending on which type we're using in the db - make sure to fix this on all the cubes

jbothma: nope! we're using bigint in the DB! interesting!

municipal_finance/update/cash_flow_v2.py
```diff
@@ -0,0 +1,95 @@
...
+        # Delete the existing matching records
+        deleted = delete_existing(
+            update_obj, CflowFactsV2, CashFlowReader, batch_size,
+        )
+        # Add the records from the dataset
+        inserted = 0
+        update_obj.file.open('r')
+        with closing(update_obj.file) as file:
+            lines = iter(file)
+            next(lines)  # Skip header
```
jbothma, line 82: could be worth adding an assertion that columns are ordered as we assume. I think we can add that to the backlog though and just document the order requirement for now

municipal_finance/update/utils.py
```diff
@@ -0,0 +1,84 @@
...
+def delete_existing(update_obj, facts_cls, reader_cls, batch_size):
+    """ Delete the existing matching records """
+    deleted = 0
+    update_obj.file.open('r')
+    with closing(update_obj.file) as file:
+        lines = iter(file)
+        next(lines)  # Skip header
+        reader = reader_cls(lines)
+        for rows in group_rows(reader, batch_size):
+            unique_keys = unique_records(rows)
```
jbothma, line 79: Am I right in thinking this is guaranteed to delete all observed demarcation code, period code pairs despite working on batches, because it might actually delete the same pair more than once, but that's ok, we're not too worried about a few duplicate deletes?

scorecard/tests/indicators/resources.py
```diff
@@ -21,28 +27,6 @@ class Meta:
         import_id_fields = ['geo_code']
 
 
-class IncexpFactsV1Resource(resources.ModelResource):
```
jbothma, line 24: did you move this out of tests because you thought initially you're able to use import-export?

.vscode/.ropeproject/config.py
juriejan: Done. Also removed the files from the repo.

municipal_finance/admin.py
juriejan: @jbothma I've added a check. I'll add a task to the backlog to reflect this behaviour on the interface.

municipal_finance/models/updates.py
juriejan: @jbothma Implemented using timestamps for file names.

municipal_finance/resources.py
juriejan: @jbothma I moved these here because they are now shared between the 'scorecard' and 'municipal_finance' apps. It feels right to me to have these defined in the application itself, as it's a utility related to the app, but I'll move it to the 'tests' directory of the 'municipal_finance' app if you prefer.

municipal_finance/settings.py
juriejan: fixed

municipal_finance/update/cash_flow_v2.py
juriejan: @jbothma Added a task to the backlog that will address checking the format and communicating errors clearly.

municipal_finance/update/utils.py
juriejan: @jbothma It will indeed sometimes attempt to delete records that have already been deleted. I used to collect all the ids that had to be deleted in a set and tried to delete them all at the same time, but it was so slow I never saw if finish (after 3 hours). I can add a quick optimisation that keeps track of what's been deleted, but I think this wouldn't improve the running time much.

scorecard/tests/indicators/resources.py
juriejan: @jbothma same answer as the previous one.

municipal_finance/resources.py
jbothma: I do prefer. It's test-only code - you can have test utility code shared between apps. Thanks

municipal_finance/update/utils.py
jbothma: no this is totally fine

municipal_finance/resources.py
juriejan: @jbothma no worries. moved.

PR #498, opened 2020-12-16
Admin update interface for fiscal and other data

juriejan:
Includes changes for the following tasks:
- Ensure that v2 data is available and working in the data explorer
- non-hardcoded cube update dates
- Rename models, tables and endpoints for v2 'bsheet' data to 'financial_position'
- Provide admin interface for updating all other data


municipal_finance/models/updates.py
```diff
@@ -2,15 +2,17 @@
 from django.db import models
 
 from django.contrib.auth.models import User
+from django.utils.deconstruct import deconstructible
 
 
-def upload_to(path):
-    def get_file_path(instance, filename):
+@deconstructible
+class UpdateFilePath(object):
```
jbothma, line 9: why not just a plain old function?

jbothma (approved):

municipal_finance/models/updates.py
juriejan: There was a point when migrations required a class to serialise field, but it's no longer needed. Thought I'd keep it that way for now till I know it's not needed again.

PR #501, opened 2021-01-27
Provide calculation sections for municipal indicators

juriejan:


assets/js/components/indicators.js
```diff
@@ -11,6 +11,17 @@ const indicatorMetricClass = {
   "": ".indicator-metric--no-status",
 };
 
+function insertSpaces(array) {
+  let result = [];
+  array.forEach((o, i, a) => {
+    result.push(o);
+    if (i !== a.length) {
```
jbothma, line 18: what are o, i and a? I don't really understand this

assets/js/components/indicators.js
```diff
@@ -56,6 +68,63 @@ export class IndicatorSection {
...
+              params['amountType'] = data.amount_type;
+            }
+            // Generate the text
+            text += `[${data.cube_name}]`;
+            text += ` item code ${data.item_codes.join(',')}`;
+            if (data.amount_type) {
+              text += `, ${data.amount_type}`;
+            }
+            params = new URLSearchParams(params).toString();
+            const url = `/api/table/${data.cube}/?${params}`;
```
jbothma, line 116:
```suggestion
            const url = `/table/${data.cube}/?${params}`;
```

because this is what a data explorer URL looks like 

https://municipaldata.treasury.gov.za/table/incexp/?year=2019&municipalities=DC10&amountType=ORGB&items=1700

assets/js/components/indicators.js
jbothma: is this an array zip or join with nbsp's? that might be simpler

jbothma (approved):
nice work!! this will be very useful

assets/js/components/indicators.js
juriejan: I've improved the code of this function, generalised it into a utility and added some documentation.

assets/js/components/indicators.js
juriejan: Made the change and included use of `API_URL` variable.

assets/js/utils.js
```diff
@@ -91,3 +91,20 @@ export function errorBoundary(f) {
...
+ * @param {*} separator - The separator that will be inserted between all the
+ *    elements in the provided array.
+ *
+ */
+export function arrayJoin(array, separator) {
+  return array.reduce((result, value, index) => {
+    return result.concat(value, separator);
+  }, []).slice(0, -1);
+}
+
```
jbothma, line 110: sorry man, I was vague before.

Isn't this what you want?

```
["bob", "dave", "lala"].join("&nbsp;")
"bob&nbsp;dave&nbsp;lala"
```

juriejan: I had to keep it in an array due to the other elements in the array being objects that are turned into elements. Essentially this is an array of HTML elements that are appended to an element in the DOM.

juriejan:
@jbothma I've made some more changes to this task yesterday since your original review. I'm providing data about cube names and amount types in the data from the view (instead of storing it in the compiled profiles) and added an initial test for these, which serves as an initial test for the profile data view.

assets/js/utils.js
jbothma: ahhh cool - I didn't read your very clear doc - sorry

PR #504, opened 2021-02-10
Bug fix: Indicator calculation links leading to 404 page

juriejan:


jbothma (approved):
looks good to me.

I've configured DATA_PORTAL_URL in prod so this should be ready to deploy.

PR #507, opened 2021-02-12
Build disestablished municipality profiles using data from before date of disestablishment

juriejan:


Reagang (approved):
Looks good. Well done

docker-compose.yml
```diff
@@ -98,7 +98,7 @@ services:
       - SENTRY_PERF_SAMPLE_RATE
       - SITE_ID=2
       - PRELOAD_CUBES=true
-      - DATA_PORTAL_URL=http://portal:8002
+      - DATA_PORTAL_URL=http://localhost:8002
```
jbothma, line 101: are you sure about this? doesn't the worker need to use the hostname `portal` to connect to the `portal` service running the api?

jbothma (approved):
awesome! looks good to me.

docker-compose.yml
juriejan: Sorry! That shouldn't have been included in this PR. Required when using Docker on MacOS, since it runs in a virtual machine, and not locally, the port needs to be exposed to be available to the browser.

PR #509, opened 2021-02-15
Bugfix: Remove default limit for this geography endpoint

Reagang:
Remove default limit for this geography endpoint by setting paginator to None and update the api call response in indicator.js

### DIFF
--- a/assets/js/components/indicators.js
+++ b/assets/js/components/indicators.js
@@ -269,7 +269,7 @@ export class IndicatorSection {
         url: '/api/geography/geography/'
       });
 
-      var activeMunicipalities = response.results.filter(function( obj ) {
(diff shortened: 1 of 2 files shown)


jbothma:
what else uses this endpoint that will break when we remove the paginator and the response structure isn't nested under `results` any more?

Reagang:
> what else uses this endpoint that will break when we remove the paginator and the response structure isn't nested under `results` any more?
mmm. You are correct. Let me have a look

jbothma (approved):

jbothma:
I agree - looks like there's nothing else using this API - I think it was added for https://municipalmoney.gov.za/infrastructure/projects/ but then the muni list was templated in server side instead.

PR #543, opened 2021-09-13
Capital project update tests

michaelglenister:
Issue: https://trello.com/c/lOpCDb44/460-ensure-that-a-project-with-precisely-the-same-details-as-the-composite-key-updates-an-existing-project-tb-2-hrs
and https://trello.com/c/9MzFnaXI/461-ensure-that-a-project-with-same-muni-and-project-number-but-different-description-results-in-a-new-project-being-created-tb-1-hr


infrastructure/tests/test_data_import.py
```diff
@@ -124,3 +126,63 @@ def test_year_error(self):
...
+        self.assertEquals(project.own_strategic_objectives, "OWN MUNICIPAL STRATEGIC OBJECTIVE")
+        self.assertEquals(project.asset_class, "Furniture and Office Equipment")
+        self.assertEquals(project.asset_subclass, "")
+        self.assertEquals(project.ward_location, "Administrative or Head Office")
+        self.assertEquals(project.longitude, 0.0)
+        self.assertEquals(project.latitude, 0.0)
+
+        self.assertEquals(Expenditure.objects.all().count(), 9)
+        expenditure = Expenditure.objects.first()
+        self.assertEquals(str(expenditure.project), " - P-CNIN FURN & OFF EQUIP")
```
jbothma, line 154:
```suggestion
        self.assertEquals(expenditure.project.description, " - P-CNIN FURN & OFF EQUIP")
```
 would be more appropriate than str() because __str__ isn't what we're trying ot test here and makes the test more brittle

infrastructure/tests/test_data_import.py
```diff
@@ -124,3 +126,63 @@ def test_year_error(self):
...
+        self.assertEquals(str(expenditure.budget_phase), "Audited Outcome")
```
jbothma, line 155:
```suggestion
        self.assertEquals(expenditure.budget_phase.name, "Audited Outcome")
```
would be more appropriate. See comment above.

infrastructure/tests/test_data_import.py
```diff
@@ -124,3 +126,63 @@ def test_year_error(self):
         self.assertEquals(AnnualSpendFile.objects.all().count(), 1)
         spend_file = AnnualSpendFile.objects.first()
         self.assertEquals(spend_file.status, AnnualSpendFile.ERROR)
+
+
+    def test_upload_project(self):
+        """Scope of Test: With no existing projects run an upload and check that the correct fields are populated"""
+        self.assertEquals(BudgetPhase.objects.all().count(), 5)
+
+        with open('infrastructure/tests/test_files/test.xlsx', 'rb', ) as f:
```
jbothma, line 135: let's test this at a lower level to close in on the unit under test.
that also allows us to use a data structure where we can more easily see the data used in the test, and when it changes, what the change was. While excel doesn't offer diffing like plain text does.

infrastructure/tests/test_data_import.py
```diff
@@ -124,3 +126,63 @@ def test_year_error(self):
...
+        self.assertEquals(Expenditure.objects.all().count(), 9)
+        expenditure = Expenditure.objects.first()
+        self.assertEquals(str(expenditure.project), " - P-CNIN FURN & OFF EQUIP")
+        self.assertEquals(str(expenditure.budget_phase), "Audited Outcome")
+        self.assertEquals(str(expenditure.financial_year), "2017/2018")
+        self.assertEquals(expenditure.amount, 340609.00)
+
+
+    def test_update_project(self):
+        """Scope of Test: With existing projects run an upload and check that the correct fields are updated"""
```
jbothma, line 161: It sounds like what we're testing is

- new projects get added
- existing projects get updated
- fields existing projects are matched by ....? fields

infrastructure/tests/test_data_import.py
```diff
@@ -124,3 +126,63 @@ def test_year_error(self):
...
+        self.assertEquals(project.iudf, "Growth")
+        self.assertEquals(project.own_strategic_objectives, "OWN MUNICIPAL STRATEGIC OBJECTIVE")
+        self.assertEquals(project.asset_class, "Furniture and Office Equipment")
+        self.assertEquals(project.asset_subclass, "")
+        self.assertEquals(project.ward_location, "Administrative or Head Office")
+        self.assertEquals(project.longitude, 0.0)
+        self.assertEquals(project.latitude, 0.0)
+
+        self.assertEquals(Expenditure.objects.all().count(), 9)
+        expenditure = Expenditure.objects.first()
```
jbothma, line 153: first only makes sense if we're asserting there's exactly 1 object. If there are more than one object we ought to filter

infrastructure/tests/test_data_import.py
michaelglenister: Using full model names, not str

infrastructure/tests/test_data_import.py
michaelglenister: As above no longer using str

infrastructure/tests/test_data_import.py
michaelglenister: Added mock data and using the lower level load_file instead of load_excel

infrastructure/tests/test_data_import.py
michaelglenister: Made test scopes more explicit

infrastructure/tests/test_data_import.py
michaelglenister: Changed to only use one project and also using .get to query specific project entries and thereby assert that they exist in the first place

infrastructure/tests/test_data_import.py
```diff
@@ -6,14 +6,76 @@
...
 
-from infrastructure.models import FinancialYear, QuarterlySpendFile, AnnualSpendFile, Expenditure, Project
+from infrastructure.models import FinancialYear, QuarterlySpendFile, AnnualSpendFile, Expenditure, Project, BudgetPhase
 from infrastructure.tests import utils
 from infrastructure.utils import load_excel
 from infrastructure.upload import process_document
 from scorecard.models import Geography
 
 
+def generate_mock_data():
```
jbothma, line 18:
```suggestion
def generate_mock_project():
```

the word "data" almost never contributes anything in a variable name.

infrastructure/tests/test_data_import.py
```diff
@@ -6,14 +6,76 @@
...
+        'GPS Longitude': '0', 'GPS Latitude': '0',
+        'Audited Outcome 2017/18': 340609.0,
+        'Full Year Forecast 2018/19': 651391.0,
+        'Budget year 2019/20': 500000.0,
+        'Budget year 2020/21': 500000.0,
+        'Budget year 2021/22': 500000.0
+    }
+    yield mock_data
+
+def mock_data_existing_project():
```
jbothma, line 38:
```suggestion
def mock_data_existing_project():
    """It is important that mock_data_new_project is kept in sync with this function"""
```

infrastructure/tests/test_data_import.py
```diff
@@ -6,14 +6,76 @@
...
+        'GPS Longitude': '1', 'GPS Latitude': '2',
+        'Audited Outcome 2018/19': 440609.0,
+        'Full Year Forecast 2019/20': 551391.0,
+        'Budget year 2020/21': 500000.0,
+        'Budget year 2021/22': 500000.0,
+        'Budget year 2022/23': 400000.0
+    }
+    yield mock_data
+
+def mock_data_new_project():
```
jbothma, line 58:
```suggestion
def mock_data_update_row():
    """Same Function, description and Number as in mock_data_existing project, different everything else"""
```

infrastructure/tests/test_data_import.py
```diff
@@ -124,3 +187,111 @@ def test_year_error(self):
...
+
+        self.assertEquals(Expenditure.objects.all().count(), 5)
+        expenditure = Expenditure.objects.get(amount=340609.00)
+        self.assertEquals(expenditure.project.project_description, "P-CNIN FURN & OFF EQUIP")
+        self.assertEquals(expenditure.budget_phase.name, "Audited Outcome")
+        self.assertEquals(expenditure.financial_year.budget_year, "2017/2018")
+
+
+    def test_update_project(self):
+        """Scope of Test: With an existing project import a new project with the same composite key and check that default fields are updated"""
```
jbothma, line 220:
```suggestion
        """Scope of Test: With an existing project import a new project with the same composite key and check that non-key fields are updated"""
```

jbothma: the term "default" doesn't make sense here - it's an implementation detail (just happens to be what create_or_update calls it)

infrastructure/tests/test_data_import.py
```diff
@@ -124,3 +187,111 @@ def test_year_error(self):
         self.assertEquals(AnnualSpendFile.objects.all().count(), 1)
         spend_file = AnnualSpendFile.objects.first()
         self.assertEquals(spend_file.status, AnnualSpendFile.ERROR)
+
+
+    def test_upload_project(self):
+        """Scope of Test: With no existing projects run an upload and check that the correct fields are populated"""
+        self.assertEquals(BudgetPhase.objects.all().count(), 5)
```
jbothma, line 194: what's this for?

infrastructure/tests/test_data_import.py
```diff
@@ -124,3 +187,111 @@ def test_year_error(self):
...
+        self.assertEquals(project.own_strategic_objectives, "OWN MUNICIPAL STRATEGIC OBJECTIVE")
+        self.assertEquals(project.asset_class, "Furniture and Office Equipment")
+        self.assertEquals(project.asset_subclass, "")
+        self.assertEquals(project.ward_location, "Administrative or Head Office")
+        self.assertEquals(project.longitude, 0.0)
+        self.assertEquals(project.latitude, 0.0)
+
+        self.assertEquals(Expenditure.objects.all().count(), 5)
+        expenditure = Expenditure.objects.get(amount=340609.00)
+        self.assertEquals(expenditure.project.project_description, "P-CNIN FURN & OFF EQUIP")
```
jbothma, line 214: If you do expenditure = project.expenditure.filter(...) you can skip the expenditure.project.project_description check, right?

infrastructure/tests/test_data_import.py
```diff
@@ -124,3 +187,111 @@ def test_year_error(self):
...
+        self.assertEquals(expenditure.budget_phase.name, "Audited Outcome")
+        self.assertEquals(expenditure.financial_year.budget_year, "2017/2018")
```
jbothma, line 216: I think we need to be checking each year and phase here to assert that all of them are created

jbothma: when we do quarterly, we'll assert exactly which ones are created, and assert that no more than the ones checked are created, thus demonstrating that only a select set are created and specific ones are ignored without listing them explicitly.

infrastructure/tests/test_data_import.py
```diff
@@ -124,3 +187,111 @@ def test_year_error(self):
...
+        self.assertEquals(project.asset_subclass, "Equipment")
+        self.assertEquals(project.ward_location, "Head Office")
+        self.assertEquals(project.longitude, 1.0)
+        self.assertEquals(project.latitude, 2.0)
+
+        self.assertEquals(Expenditure.objects.all().count(), 8)
+        expenditure = Expenditure.objects.get(amount=440609.00)
+        self.assertEquals(expenditure.project.project_description, "P-CNIN FURN & OFF EQUIP")
+        self.assertEquals(expenditure.budget_phase.name, "Audited Outcome")
+        self.assertEquals(expenditure.financial_year.budget_year, "2018/2019")
```
jbothma, line 257: same as above
- we wanna be sure that the expenditure items are associated wit the project, and we can do that via foreign key
- we wanna be sure that the ones that exist (that were checked above) had their amount updated, and the new ones were created (like audit outcome here)

jbothma (changes requested):
we're basically there now. Just a few small comments that should be quick to fix.

infrastructure/tests/test_data_import.py
michaelglenister: What I tried to do is check that the expected amount of BudgetPhases are created - to prevent cases like that bug where a budget column was not being created/updated for Full Year Forecast

infrastructure/tests/test_data_import.py
michaelglenister: I think the expenditure.project.project_description check was not necessary here

infrastructure/tests/test_data_import.py
michaelglenister: I made some changes to increase coverage of relevant expenditures, any expenditure that should update is checked

infrastructure/tests/test_data_import.py
michaelglenister: test_new_project() now uses foreign keys to confirm new expenditures are associated with the new project. For other tests, there is only one project so all the project_id foreign keys for expenditures are the same

infrastructure/tests/test_data_import.py
jbothma: It's not obvious why this is relevant to the test - why the number of budget phases in the db before the test is relevant to the test. Perhaps a comment would help. I still don't totally understand.

infrastructure/tests/test_data_import.py
jbothma: I thought the idea was to assert that this expenditure item is related to the project. It's fine as is as long as expenditure items must always be associated to one project (not null) and this test asserts that there is exactly one project in the db

infrastructure/tests/test_data_import.py
```diff
@@ -6,14 +6,85 @@
...
+        'Budget year 2021/22': 5000.0,
+        'Budget year 2022/23': 6100.0
+    }
+    yield mock_data
+
+
+def verify_expenditure(this, amount, budget_phase, year):
+    expenditure = Expenditure.objects.get(amount=amount)
+    this.assertEquals(expenditure.budget_phase.name, budget_phase)
+    this.assertEquals(expenditure.financial_year.budget_year, year)
```
jbothma, line 85: I'd call this `verify_expenditure_exists` otherwise I have to read the method definition to know what it does.

I'd also include the other fields as get() parameters, otherwise this will fail when more than one expenditure item have the same amount. That's future-proofing and not super urgent now, but an easy improvement.

infrastructure/tests/test_data_import.py
```diff
@@ -124,3 +196,114 @@ def test_year_error(self):
...
+        self.assertEquals(project.asset_subclass, "Equipment")
+        self.assertEquals(project.ward_location, "Head Office")
+        self.assertEquals(project.longitude, 1.0)
+        self.assertEquals(project.latitude, 2.0)
+
+        self.assertEquals(Expenditure.objects.all().count(), 10)
+
+        expenditure_id = Expenditure.objects.get(amount=2100).project_id
+        new_expenditures = Expenditure.objects.filter(project_id=expenditure_id)
+        self.assertEquals(new_expenditures.count(), 5)
```
jbothma, line 305: This would have been more readable if you named the original project initial_project and the second project new_project and then did 
```
         new_expenditures = Expenditure.objects.filter(project=new_project)
```

infrastructure/tests/test_data_import.py
```diff
@@ -6,14 +6,85 @@
...
+        'Audited Outcome 2018/19': 2100.0,
+        'Full Year Forecast 2019/20': 3100.0,
+        'Budget year 2020/21': 4000.0,
+        'Budget year 2021/22': 5000.0,
+        'Budget year 2022/23': 6100.0
+    }
+    yield mock_data
+
+
+def verify_expenditure(this, amount, budget_phase, year):
```
jbothma, line 82: I think calling it `this` or `self` is a bit weird when this is a function totally outside of the class. Let's rather refer to it as `test_case` so it's very clear what it's supposed to be an instance of.

```suggestion
def verify_expenditure(testcase, amount, budget_phase, year):
```

jbothma (approved):

PR #544, opened 2021-09-16
Updates to project detail page 

michaelglenister:
Issue: https://trello.com/c/SdzTftO4/462-project-detail-page-ensure-that-project-detail-page-shows-the-latest-data-available-for-a-given-project


infrastructure/migrations/0019_project_latest_implementation_year.py
```diff
@@ -0,0 +1,22 @@
...
+        ('infrastructure', '0018_auto_20210915_1645'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='project',
+            name='latest_implementation_year',
+            field=models.ForeignKey(default=infrastructure.models.get_finanial_year_default, on_delete=django.db.models.deletion.CASCADE, to='infrastructure.FinancialYear'),
+        ),
+    ]
```
jbothma, line 22: A better approach than providing a default method:

- add the field with null=True
- run some python code which, if projects exist, fetches the year 2019/20 (assume it exists), and update the projects with that year selected
- change the field to null=False

michaelglenister: Reworked the migrations further and included this change

PR #547, opened 2021-10-05
Setting up Selenium tests

michaelglenister:
Issue: https://trello.com/c/SdzTftO4/462-project-detail-page-ensure-that-project-detail-page-shows-the-latest-data-available-for-a-given-project


infrastructure/migrations/0018_auto_20210928_1642.py
```diff
@@ -0,0 +1,39 @@
...
+def populate_implementation_year(apps, schema_editor):
+    project = apps.get_model("infrastructure", "Project")
+    year = apps.get_model("infrastructure", "FinancialYear")
+
+    db_alias = schema_editor.connection.alias
+
+    try:
+        implementation_year_id = year.objects.using(db_alias).get(budget_year="2019/2020")
+    except ObjectDoesNotExist:
+        implementation_year_id = 1
```
jbothma, line 19: we should be able to merge master into this branch and this should change to the fixed implementation, right?

infrastructure/tests/helpers.py
```diff
@@ -0,0 +1,46 @@
...
+        budget_phase = BudgetPhase.objects.get(name=phase)
+        financial_year = FinancialYear.objects.get_or_create(budget_year=year)
+
+        expenditure = Expenditure.objects.create(
+            project=self.project,
+            budget_phase=budget_phase,
+            financial_year=financial_year[0],
+            amount=amount,
+        )
+        return expenditure
```
jbothma, line 46: this doesn't really feel like it belongs as part of a BaseSeleniumTestCase, does it? If it does, this class just becomes a bag of "whatever we feel like chucking into the class".

looking at the method, it assumes there exists a self.project, which is a very big assumption specific to a particular test.

I think this method should rather be somewhere outside of this class, letting this class focus on supporting selenium tests. I don't see a strong argument for this to be part of any class, to be honest. It could just be a method on its own in the test module as far as I can tell.

municipal_finance/middleware.py
```diff
@@ -38,7 +38,7 @@ def __call__(self, request):
     
     def process_request(self, request):
         site = get_current_site(request)
-        if site.name == 'Scorecard':
+        if site.name == 'Scorecard' or site.name == 'example.com':
```
jbothma, line 41: could you see which site objects exist in the DB at this point? Our migrations should be creating the right ones. Perhaps we need the test to edit the relevant object to have the netloc (hostname and port) used in the test?

infrastructure/migrations/0018_auto_20210928_1642.py
michaelglenister: Yup, merged

infrastructure/tests/helpers.py
michaelglenister: Agreed, there's no benefit to having that method clutter helpers/utils unless it's shared by other modules

municipal_finance/middleware.py
michaelglenister: For some reason one of the selenium tests reverts to the default example.com domain. I removed the code changes to prod and update the domain in the test

PR #548, opened 2021-10-13
Filter projects by latest implementation year

michaelglenister:
Issue: https://trello.com/c/BXAe34Nf/463-ensure-that-a-user-is-able-to-find-latest-implementation-year-projects-in-search-results


scorecard/views.py
```diff
@@ -176,26 +177,37 @@ def get_context_data(self, *args, **kwargs):
...
-        page_json["infrastructure_summary"] = {
-            "projects": [infra_dict(p) for p in infrastructure[:5]],
-            "project_count": infrastructure.count(),
-            "financial_year": infrastructure_financial_year[5:9]
-        }
+
+            page_json["infrastructure_summary"] = {
+                "projects": [infra_dict(p) for p in infrastructure[:5]],
+                "project_count": infrastructure.count(),
+                "financial_year": financial_year.budget_year[5:9]
```
jbothma, line 204: what are these magic numbers? how do they work?

scorecard/views.py
```diff
@@ -176,26 +177,37 @@ def get_context_data(self, *args, **kwargs):
...
-                "geography",
-                "expenditure__budget_phase",
-                "expenditure__financial_year",
-                "expenditure",
-            )
-            .filter(
-                geography__geo_code=self.geo_code,
-                expenditure__budget_phase__name="Budget year",
-                expenditure__financial_year__budget_year=infrastructure_financial_year,
+        if Project.objects.all().count() > 0:
```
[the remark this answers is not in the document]
michaelglenister: It looks like the reason I added this was because local didn't work when there were no capital projects, I don't think that will ever be the case in prod. Will remove this

infrastructure/views/api.py
```diff
@@ -153,16 +153,15 @@ def text_search(self, qs, text):
...
+        budget_phase = params.get("budget_phase")
 
         return {"total": qs.total_value(financial_year, budget_phase)}
 
     def add_filters(self, qs, params):
-        """
-        Going to hard code params for budget phase and financial year,
-        """
+        if 'financial_year' in params.keys():
+            qs = qs.filter(latest_implementation_year__budget_year=params['financial_year'])
```
jbothma, line 163: why does this need to change?

municipal_finance/settings.py
```diff
@@ -153,6 +153,12 @@
         "the selected year and quarter.",
         "quarter_select",
     ],
+    "CAPITAL_PROJECT_SUMMARY_YEAR": [
+        "2019/2020",
+        "The year to use when fitlering which capital projects to  which grant spending data is available. "
```
jbothma, line 158: copy pasta "grant spending"

seeddata.json
```diff
@@ -38,5 +38,13 @@
     "code": "",
     "name": "Adjusted Budget"
   }
+},
+{
+  "model": "infrastructure.financialyear",
+  "pk": 1,
+  "fields": {
+    "budget_year": "2019/2020",
+    "active": "True"
```
jbothma, line 47: this PR should mean that we can remove the active property, right?

infrastructure/tests/test_data_import.py
```diff
@@ -143,7 +143,7 @@ def test_file_upload(self):
     def test_file_upload_fail(self):
         """Scope of Test: Testing if the file upload fail in Django Admin to processing file and add to Django_Q"""
         self.client.login(username=self.username, password=self.password)
-        fy = FinancialYear.objects.create(budget_year="2019/2020", active=1)
+        fy, _ = FinancialYear.objects.get_or_create(budget_year="2019/2020")
```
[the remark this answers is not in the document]
michaelglenister: Agreed. I actually think it's important to fix this because an ambiguousness like this could make the test pass even though the code is broken. Changing to use get() because we expect a year to exist

infrastructure/views/api.py
michaelglenister: This filters projects in the search according to the financial period. Changing this to use Constance config, instead of a defensive if statement

seeddata.json
michaelglenister: Yes, will remove that from seed data

municipal_finance/settings.py
michaelglenister: Woops, will change that

infrastructure/views/api.py
```diff
@@ -108,12 +107,13 @@ class ProjectSearch(generics.ListCreateAPIView):
...
+        "geography__name": "municipality",
         "function": "function",
         "project_type": "project_type",
-        "province": "geography__province_name",
-        "budget_phase": "expenditure__budget_phase__name",
-        "financial_year": "expenditure__financial_year__budget_year",
+        "geography__province_name": "province",
+        "expenditure__budget_phase__name": "budget_phase",
+        "expenditure__financial_year__budget_year": "financial_year",
+        "latest_implementation_year__budget_year": "financial_year",
```
jbothma, line 116: are we supposed to have both of these? it looks like this means we'll do

```
qs.filter(
    expenditure__financial_year__budget_year=qsparam, 
    latest_implementation_year__budget_year=qsparam
)
```

which means we'll only get projects that both match the provided financial year in their latest_implemetation_year and also has an expenditure item for that year. we don't care about the latter though.

jbothma:
```suggestion
        "latest_implementation_year__budget_year": "financial_year",
```

michaelglenister: I agree but I think this is outside the scope of this PR which is to be able to filter projects by a configurable latest implementation year. This should be fixed in #553 where annual and quarterly uploads should display projects that don't necessarily have a 'Budget year' expenditure for the latest implementation year

PR #550, opened 2021-11-02
Ensure quarterly budget uploads can create and update capital projects

michaelglenister:
- Show notice if this project is not current (implementation year is older than the current summary year )
- Fix quarter creation based on column heading parsing
- add tests


infrastructure/static/js/mm-webflow.js
```diff
@@ -844,6 +848,11 @@ function mmWebflow(js) {
 
         $(".detail-button_wrapper").hide();
         $(".subsection-chart__detail").hide();
+
+        var summaryYear = js.summary_year
+        if (summaryYear.split('/')[0] > implementYear.split('/')[0]) {
+            $(".project-details__info-message").parent().append('This project was last updated with ' + implementYear + ' data. Please see the search page for the latest figures.');
```
jbothma, line 854: this project _should_ be updated with the latest figures if this project exists in later years, right?

So the message should perhaps refer to "latest projects" rather than "latest figures"

jbothma: nice thinking though!

municipal_finance/middleware.py
```diff
@@ -38,7 +38,7 @@ def __call__(self, request):
     
     def process_request(self, request):
         site = get_current_site(request)
-        if site.name == 'Scorecard':
+        if site.name == 'Scorecard' or site.name == 'example.com':
```
jbothma, line 41: shouldn't be here

scorecard/views.py
```diff
@@ -176,26 +177,37 @@ def get_context_data(self, *args, **kwargs):
...
-        }
+
+            page_json["infrastructure_summary"] = {
+                "projects": [infra_dict(p) for p in infrastructure[:5]],
+                "project_count": infrastructure.count(),
+                "financial_year": financial_year.budget_year[5:9]
+            }
+        else:
+            page_json["infrastructure_summary"] = {
+                "projects": "",
```
jbothma, line 208: watch out str instead of list

scorecard/views.py
```diff
@@ -176,26 +177,37 @@ def get_context_data(self, *args, **kwargs):
...
-                "geography",
-                "expenditure__budget_phase",
-                "expenditure__financial_year",
-                "expenditure",
-            )
-            .filter(
-                geography__geo_code=self.geo_code,
-                expenditure__budget_phase__name="Budget year",
-                expenditure__financial_year__budget_year=infrastructure_financial_year,
+        if Project.objects.all().count() > 0:
```
jbothma, line 180: irrelevant query

infrastructure/static/js/mm-webflow.js
michaelglenister: Resolving comments that were addressed to tidy this PR

infrastructure/static/js/mm-webflow.js
```diff
@@ -847,6 +847,11 @@ function mmWebflow(js) {
 
         $(".detail-button_wrapper").hide();
         $(".subsection-chart__detail").hide();
+
+        var summaryYear = js.summary_year
+        if (summaryYear.split('/')[0] > implementYear.split('/')[0]) {
```
jbothma, line 852: you don't need the split to do a valid comparison, do you?

michaelglenister: Interesting, so 2019/2020 is run as a division operation. But that should be okay because successive periods will always be slightly greater than previous ones. Removed the splits

jbothma (approved):
thanks! ship it.

PR #557, opened 2021-12-14
Make it possible to create and display notices on the site pages

michaelglenister:


jbothma:
please also show this notice in admin and on infra project pages - those could be separate pull requests if you wanna get this out the door sooner

jbothma (approved):

PR #558, opened 2022-01-05
Prevent search engine indexing

michaelglenister:
Added environment and context variables
Updated the webflow import script and webflow pages

According to Google docs, this should stop indexing by most search engines

Note this requires the NO_INDEX env variable to be set to prevent indexing


fabform:
Handy alternative list of [alternative search engines](https://fabform.io/a/alternative-search-engines)

jbothma (approved):
Please add a test for this (at least one case where it not being set results in the tag not being present, otherwise similar to the notices test should be fine, right?)

Then ship it

PR #559, opened 2022-01-14
Show which formula version is used on cube data

michaelglenister:


jbothma:
merging all this via https://github.com/OpenUpSA/municipal-data/pull/564

PR #570, opened 2022-03-08
Add noindex meta tag to infrastructure pages

michaelglenister:
https://trello.com/c/EiaJsovr/467-exclude-staging-site-from-search-engines


jbothma (approved):
I'm beginning to think we need to include these in the automated tests :(

I understand now why django has https://docs.djangoproject.com/en/4.0/topics/testing/tools/#django.test.SimpleTestCase.assertTemplateUsed
You might then only assert that the base template was used, implement this on the base template, and test this functionality in one place.

Another vote against webflow exports. 

Could you add this to the tests and then ship?

PR #573, opened 2022-03-17
Get only distinct results from infrastructure project seach filters

michaelglenister:
Issue: https://trello.com/c/Pt7UjTND/508-duplicate-projects-and-unexpected-results-when-filtering-and-searching-infrastructure-projects

dupimage

### DIFF
--- a/infrastructure/tests/test_api_views.py
+++ b/infrastructure/tests/test_api_views.py
@@ -1,30 +1,28 @@
-from django.test import Client, TestCase
-from infrastructure import utils
-from infrastructure import models
-import json
-
-from infrastructure.models import FinancialYear, QuarterlySpendFile, Expenditure, Project
-
-from scorecard.models import Geography
-from scorecard.profiles import MunicipalityProfile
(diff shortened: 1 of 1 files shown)


infrastructure/views/api.py
```diff
@@ -175,7 +178,7 @@ def add_filters(self, qs, params, filter_map):
             if v in params:
                 query_dict[k] = params[v]
 
-        return qs.filter(**query_dict)
+        return qs.filter(**query_dict).distinct()
```
jbothma, line 181: Is there test coverage for this?

Happy to ship this change first and ship the test as a second PR.

jbothma (approved):

infrastructure/views/api.py
michaelglenister: Thanks. there's a wide reaching, end to end test called test_search_with_quarterly that should cover this. The problem was caused by joining the two queries into one and not filtering by distinct. That test will complain if that changes

Yes could add a unit test in addition to that

PR #577, opened 2022-04-11
Show 'no report available' message if opinion report URL is null

michaelglenister:
Issue: https://trello.com/c/puyizh9F/519-audit-report-years-config-why-is-audit-report-2015-16-blank-for-buf-on-staging-are-we-doing-the-wrong-thing-with-mapping-years-t

### DIFF
--- a/assets/js/components/audit-opinions.js
+++ b/assets/js/components/audit-opinions.js
@@ -14,10 +14,10 @@ class Icon {
 }
 
 class ReportCard {
-  constructor($element, report) {
+  constructor($element, key, report) {
     this.$element = $element;
     logIfUnequal(1, this.$element.length);
-    this.$element.find(".audit-outcome__year").text(formatFinancialYear(report.date));
+    this.$element.find(".audit-outcome__year").text(formatFinancialYear(key));
     this.$element.find(".audit-outcome__heading").text(report.result);
 
     if (report.report_url === null) {
@@ -60,10 +60,12 @@ export class AuditOpinions {
   constructor(reports) {
     this.$element = $(this.selector);
     logIfUnequal(1, this.$element.length);
-
-    reports.values.forEach((report, index) => {
+    // Using Object.entries() and sort() as documenteded here
+    // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries
+    // to loop through the reports["values"] objects and order by key
+    Object.entries(reports["values"]).sort((a, b) => b[0].localeCompare(a[0])).forEach(([key, value], index) => {
       const $element = this.$element.find(`.audit-outcome:eq(${ index })`);
-      new ReportCard($element, report);
+      new ReportCard($element, key, value);
     });
   }
 }
--- a/assets/js/components/audit-opinions.js
+++ b/assets/js/components/audit-opinions.js
@@ -20,16 +20,17 @@ class ReportCard {
     this.$element.find(".audit-outcome__year").text(formatFinancialYear(key));
     this.$element.find(".audit-outcome__heading").text(report.result);
 
-    if (report.report_url === null) {
+    let reportURL = report.report_url;
+    if (reportURL === null) {
       this.$element.find(".audit-outcome__download").text("No report available");
     }
-    else if (report.report_url.endsWith('.pdf')) {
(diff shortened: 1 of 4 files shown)


assets/js/components/audit-opinions.js
```diff
@@ -20,14 +20,14 @@ class ReportCard {
...
+      this.$element.find(".audit-outcome__download").text("No report available");
+    }
+    else if (report.report_url.endsWith('.pdf')) {
       report.report_url = report.report_url.substring(0, report.report_url.lastIndexOf("/"));
     }
 
     this.$element.find(".audit-outcome__download").attr("href", report.report_url);
-    if (report.report_url === null) {
-      this.$element.find(".audit-outcome__download").text("No report available");
-    }
```
jbothma, line 30: I see now this is modifying someone else's data. Similarly to https://github.com/OpenUpSA/municipal-data/pull/578/files#r853249892, that could have unexpected consequences. Better style would be to create a local variable that is report.report_url if it doesn't end with .pdf, otherwise we strip the pdf and assign that. And then use the local variable for the PDF URL.

assets/js/components/audit-opinions.js
```diff
@@ -20,14 +20,14 @@ class ReportCard {
...
 
-    if (report.report_url.endsWith('.pdf')) {
+    if (report.report_url === null) {
+      this.$element.find(".audit-outcome__download").text("No report available");
+    }
+    else if (report.report_url.endsWith('.pdf')) {
       report.report_url = report.report_url.substring(0, report.report_url.lastIndexOf("/"));
     }
 
     this.$element.find(".audit-outcome__download").attr("href", report.report_url);
```
jbothma, line 30: I think this needs to be included in the "if null" else clause, because it doesn't make sense to set href=null, right?

scorecard/profile_data/api_data.py
```diff
@@ -178,17 +178,33 @@ def muni_contact(self):
...
+            if values:
+                if values[-1]["date"] != year:
+                    values.append(
+                        {
+                            "date": year,
+                            "result": None,
+                            "rating": None,
+                            "report_url": None,
+                        }
+                    )
```
jbothma, line 206: this is pretty clunky. This is saying "if we find the year, then add it to the list of years. and if the last year we added isn't the current year, then add a placeholder year"

I think what would make more sense is

1. loop over the data you have. for each item, add it to a dict keyed on year
2. loop over the years you expect.
  - if you have it in your dict, add it to your output list
  - if you don't, add a placeholder to your output list

assets/js/components/audit-opinions.js
michaelglenister: Added changes to address this

assets/js/components/audit-opinions.js
michaelglenister: Changed to not set href to null

scorecard/profile_data/api_data.py
michaelglenister: What made this messy was that audit opinions were stored as a list of dictionaries. Changed that to a dictionary keyed by year

This means a profile compilation is needed for these changes to work

scorecard/profile_data/api_data.py
```diff
@@ -178,20 +178,25 @@ def muni_contact(self):
...
+                "rating": result["opinion.code"],
+                "report_url": result["opinion.report_url"],
+            }
+        for year in years:
+            if year not in opinions:
+                opinions[year] = {
+                    "result": None,
+                    "rating": None,
+                    "report_url": None,
                 }
```
jbothma, line 197:
```suggestion
        for year in years:
            if year in opinions:
                values.append(opinions[year])
            else:
                values.append({
                    "result": None,
                    "rating": None,
                    "report_url": None,
                })
        return values
```

jbothma: this should simplify the change set

jbothma: and make it consistent with how the rest of the indicators work (array of years incl placeholders for missing data)

michaelglenister: Added these changes and reverted unnecessary changes to data format and demo-data

jbothma (approved):

PR #578, opened 2022-04-12
V1 formula should link to 2019 years on the Data Explorer

michaelglenister:
2019 will be the last year V1 formulas are used

Issue: https://trello.com/c/ufXhca2c/530-v1-formula-should-link-to-latest-v1-year-hardcoding-2018-19-is-fine


assets/js/components/indicators.js
```diff
@@ -116,6 +116,7 @@ export class IndicatorSection {
       if ($textEl.length && textData) {
           $textEl.text(textData);
       }
+      this.sectionData.last_year = "2019"
       this.setFormulaLink(formulaData, ".is--pre-2019-20")
```
jbothma, line 120: I don't think changing `sectionData` is the right approach. We'd be modifying what the backend told us is the data for this indicator.

Perhaps passing a year to `setFormulaLink` is more appropariate? For the v2 link we can pass `this.sectionData.last_year` and for the v1 link we can pass `"2019"`.

michaelglenister: Changed

jbothma (approved):

PR #587, opened 2022-05-31
Disaggregate capex types

michaelglenister:
Issue: https://trello.com/c/qnnz0mt1/457-disaggregate-capex-types-as-a-user-i-would-liketo-be-able-to-see-different-types-of-capital-acquisition-amounts-distinguish-new


models/capital_v2.json
```diff
@@ -157,5 +157,28 @@
             "column": "amount",
             "label": "Amount"
         }
+    },
+    "can_aggregate": "True",
```
jbothma, line 161: actually it's that it can't aggregate - can_aggregate should be false, and it should be a boolean, not a string.

Also it's not the cube that can/can't aggregate, it's the dimension. So the right place for a property like this is `dimensions.capital_type`

models/capital_v2.json
```diff
@@ -157,5 +157,28 @@
...
+            "label": "Renewal"
+        },
+        "repairs_maintenance": {
+            "column": "REPAIR_MNT",
+            "label": "Repairs and maintenance"
+        },
+        "asset_register_summary": {
+            "column": "UPGRADING",
+            "label": "Upgrading"
+        }
```
jbothma, line 182: this seems to be hardcoding what's already available here https://municipaldata.treasury.gov.za/api/cubes/capital_v2/members/capital_type

One way to access these labels for presentation would be to perform that query. Another way would be to include the label as a drilldown as well, similarly to how we do item code and label.

municipal_finance/static/javascript/table.js
```diff
@@ -714,22 +722,31 @@
       var table = this.$('.values').empty()[0];
       var functions = this.functionHeadings();
 
+      var muniColumns = 1;
+      if (cube.model.can_aggregate) {
+        muniColumns = Object.keys(cube.model.aggregate_columns).length;
+        var aggregateColumns = cube.model.aggregate_columns;
+      }
+      else if (cube.columns.length > 1) {
```
jbothma, line 730:
```suggestion
      } else if (cube.columns.length > 1) {
```

municipal_finance/static/javascript/table.js
```diff
@@ -714,22 +722,31 @@
...
+        muniColumns = cube.columns.length;
+        var aggregateColumns = cube.model.measures;
+      }
```
jbothma, line 733: It's nicer to declare the variable once and then assign to it in the if context. It's a bit weird declaring a variable in two places, also to be declaring it in a deeper context than where you use it later.

```suggestion
      var muniColumns = 1;
      var aggregateColumns;
      if (cube.model.can_aggregate) {
        muniColumns = Object.keys(cube.model.aggregate_columns).length;
        aggregateColumns = cube.model.aggregate_columns;
      }
      else if (cube.columns.length > 1) {
        muniColumns = cube.columns.length;
        aggregateColumns = cube.model.measures;
      }
```

municipal_finance/static/javascript/table.js
```diff
@@ -741,13 +758,12 @@
       }
 
       // column (aggregate) headings
-      if (cube.columns.length > 1) {
+      if (cube.model.can_aggregate || cube.columns.length > 1) {
```
jbothma, line 761: I think let's use `if cube.name == 'capital_v2'` here instead.

The need for an additional row of column headings is a bit more complex than simply that _some_ column in the cube can not be aggregated. So let's not do logic based on can_aggregate yet.

municipal_finance/static/javascript/table.js
```diff
@@ -800,7 +816,23 @@
                 var data = muni_data ? muni_data[functions[f].code] : null;
                 this.renderMuniValues(muni, data, tr);
               }
-            } else {
+            }
+            else if (cube.model.can_aggregate) {
```
jbothma, line 820: I think let's use `if cube.name == 'capital_v2'` here instead.

The need for an additional row of column headings is a bit more complex than simply that _some_ column in the cube can not be aggregated. So let's not do logic based on can_aggregate yet.

municipal_finance/static/javascript/table.js
```diff
@@ -714,22 +722,31 @@
       var table = this.$('.values').empty()[0];
       var functions = this.functionHeadings();
 
+      var muniColumns = 1;
+      if (cube.model.can_aggregate) {
+        muniColumns = Object.keys(cube.model.aggregate_columns).length;
+        var aggregateColumns = cube.model.aggregate_columns;
```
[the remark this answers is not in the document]
michaelglenister: Changed to valueColumns

municipal_finance/static/javascript/table.js
michaelglenister: changed to `if (CUBE_NAME == 'capital_v2')`

municipal_finance/static/javascript/table.js
michaelglenister: Included this change in the latest commit

municipal_finance/static/javascript/table.js
michaelglenister: Changed all cases like this

models/capital_v2.json
michaelglenister: Now fetching the labels from that endpoint. Where` this.renderValues` is being called, the `this` object  does not exist in that function, I'm not sure why it's doing that so I just parsed the `aggregate_columns` variable directly

municipal_finance/static/javascript/table.js
```diff
@@ -800,7 +831,22 @@
...
+                      }
+                    }
+                    data.push(row);
+                });
+              }
+              this.renderMuniValues(muni, data, tr);
+            }
+            else {
               this.renderMuniValues(muni, muni_data && muni_data[0], tr);
             }
```
jbothma, line 851: The capital cubes have the functions dimension, which currently offers the user the option to disaggregate by government function.

Screenshot_2022-09-02_13-44-10

I think to get this change out the door, it's ok to disable the function controls for capital_v2 cube. That would _kind of_ make it ok to say `if functions.length OR if cube == capital_v2` because we're then ensuring that we can't have both of those bring true. Perhaps to highlight that things are getting dodgy, please change

```
if (functions.length > 0) {
```

to 

```
if (functions.length > 0 && CUBE_NAME != 'capital_v2') {
```

so this will become 

```
            if (functions.length > 0 && CUBE_NAME != 'capital_v2') {
              for (var f = 0; f < functions.length; f++) {
                var data = muni_data ? muni_data[functions[f].code] : null;
                this.renderMuniValues(muni, data, tr);
              }
            } else if (CUBE_NAME == 'capital_v2') {
```

I see that what we really want is a pivot table rendering component and we don't want to build one ourselves because it's pain.

If you want to be extra cautious, you could add a clause

```
} else if (functions.length > 0 && CUBE_NAME == 'capital_v2') {
    alert("An unexpected error occurred.");
    raise("A user shouldn't be able to disaggregate by function on the capital_v2 cube at the moment");
}
```

models/capital_v2.json
```diff
@@ -89,7 +89,8 @@
             "label_attribute": "label",
             "description": "",
             "label": "Capital Type",
-            "join_column": ["capital_type_id", "id"]
+            "join_column": ["capital_type_id", "id"],
+            "can_aggregate": false
```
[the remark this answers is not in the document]
michaelglenister: Not sure why I changed this, might be a leftover of code I didn't use. Reverting back to non_aggregatable: true

municipal_finance/static/javascript/table.js
```diff
@@ -687,6 +710,7 @@
 
       if (cube.columns.length > 1) blanks++;
       if (!_.isEmpty(this.filters.get('functions'))) blanks++;
+      if (CUBE_NAME == 'capital_v2') blanks++;
 
       for (var i = 0; i < blanks; i++) {
         var spacer = $('<th>').html('&nbsp;').addClass('spacer');
```
paulmwatson, line 716: Is there a specific reason why we'er setting all `th` to `&nbsp;` and not just empty?

municipal_finance/static/javascript/table.js
```diff
@@ -800,7 +832,22 @@
                 var data = muni_data ? muni_data[functions[f].code] : null;
                 this.renderMuniValues(muni, data, tr);
               }
-            } else {
+            } else if (CUBE_NAME == 'capital_v2') {
+              var data = [];
+              if (muni_data){
+                Object.keys(aggregate_columns).forEach(column => {
+                  var row = "";
+                    for (var f = 0; f < muni_data.length; f++) {
```
paulmwatson, line 840: Indenting is off. We should look to use a formatter rather than relying on hand formatting.

municipal_finance/static/javascript/table.js
michaelglenister: It seems to add a space to the `th`, `''` and `' '` result in the following
image

municipal_finance/static/javascript/table.js
michaelglenister: I adjusted the formatting here. I try to fix formatting in table.js where I touch code but it's probably a good idea to do one big format and then it's good

paulmwatson (approved):

PR #589, opened 2022-06-07
Changes to make the lastes audit opinion scraping work

jbothma:
-Adjusting excel sheet indexing
-Waiting for HTTP response 429 to reset, the site doesn't return a
waiting time so just wait 11 seconds
-Merge reports


municipal_finance/data_import/audit_reports.py
```diff
@@ -69,6 +70,10 @@ def expand_dir(self, dir, params):
         print("\nDir: %s %r" % (dir, params))
         try:
             r = requests.get(root + dir, params=params)
+
+            while r.status_code == 429:
```
jbothma, line 74: well done sharepoint.  :clap:

PR #595, opened 2022-07-01
Add site notices to data explorer

michaelglenister:
Issue: https://trello.com/c/cOcFTbQS/525-show-site-notices-on-data-explorer-site-incl-docs-so-that-users-who-somehow-land-there-know-that-it-isnt-official-data-est-1d

### DIFF
--- a/municipal_finance/templates/docs.html
+++ b/municipal_finance/templates/docs.html
@@ -34,8 +34,8 @@
 
 <div class="page-wrapper">
   <div class="content">
-    {% for notice in site_notices %}<h3>{{ notice.content | safe }}</h3>{% endfor %}
     <h1 id="intro">Municipal Money API Documentation</h1>
+    {% for notice in site_notices %}<div class="site-notice">{{ notice.content | safe }}</div>{% endfor %}
     <aside class="info">
       Usage of this data is subject to the <b><a href="/terms">Terms of Use</a></b>
     </aside>
--- a/municipal_finance/templates/docs.html
+++ b/municipal_finance/templates/docs.html
@@ -34,6 +34,7 @@
 
 <div class="page-wrapper">
   <div class="content">
+    {% for notice in site_notices %}<h3>{{ notice.content | safe }}</h3>{% endfor %}
     <h1 id="intro">Municipal Money API Documentation</h1>
     <aside class="info">
       Usage of this data is subject to the <b><a href="/terms">Terms of Use</a></b>
--- a/municipal_finance/templates/layout_data.html
+++ b/municipal_finance/templates/layout_data.html
@@ -16,8 +16,9 @@
     <h1><a href="/">Municipal Finance Data</a></h1>
   </div>
 </header>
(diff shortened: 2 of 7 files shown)


municipal_finance/templates/table.html
```diff
@@ -23,7 +23,12 @@
     </div>
 
     <img src="{% static 'images/treasury-logo.png' %}" class="logo">
-    <h1><a href="/">Municipal Finance Data Tables</a></h1>
+    {% if site_notices %}
+      <h3><a href="/">Municipal Finance Data Tables</a></h3>
```
paulmwatson, line 27: This seems to be the top level heading on the page so it should use h1 not change to h3. Note CSS h3 styling above too.

municipal_finance/templates/docs.html
```diff
@@ -34,6 +34,7 @@
 
 <div class="page-wrapper">
   <div class="content">
+    {% for notice in site_notices %}<h3>{{ notice.content | safe }}</h3>{% endfor %}
```
paulmwatson, line 37: Let's not use a h3 for notices as they are not structurally headings. A p with a class should suffice e.g. `<p class="site-notice">....</p>`

municipal_finance/templates/table.html
michaelglenister: Changed the heading tag and corresponding CSS

municipal_finance/templates/docs.html
michaelglenister: Same here, changed to div with it's own CSS

municipal_finance/templates/layout_data.html
```diff
@@ -16,6 +16,7 @@
     <h1><a href="/">Municipal Finance Data</a></h1>
   </div>
 </header>
+{% for notice in site_notices %}<h4>{{ notice.content | safe }}</h4>{% endfor %}
```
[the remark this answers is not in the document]
michaelglenister: Changed this too and CSS

municipal_finance/templates/table.html
```diff
@@ -23,7 +23,13 @@
     </div>
 
     <img src="{% static 'images/treasury-logo.png' %}" class="logo">
-    <h1><a href="/">Municipal Finance Data Tables</a></h1>
+    {% if site_notices %}
+        <h1 class="site-notice"><a href="/">Municipal Finance Data Tables</a></h1>
+        {% for notice in site_notices %}<h1 class="site-notice">{{ notice.content | safe }}</h1>{% endfor %}
```
paulmwatson, line 28: `h1` used for site notice here, other changes are to use a `div`.

paulmwatson:
Also looking at the code again I think the repeated blocks of `{% for notice in site_notices %}` can be DRYed up into a single file/template to include wherever `site-notice` is needed. I'd also like to see the CSS specified just once, not sure why it needs different styling for different places. Is the current code base a bit inconsistent in styling?

PR #610, opened 2022-08-23
Create and populate primary keys as an int

michaelglenister:


municipal_finance/migrations/0041_create_new_key_columns.py
```diff
@@ -0,0 +1,56 @@
...
+        ),
+        migrations.RunPython(populate_items_id),
+        migrations.AlterField(
+            model_name='agedcreditorfactsv1',
+            name='item_code',
+            field=models.TextField(),
+        ),
+        #migrations.RunSQL(
+        #    ['ALTER TABLE aged_creditor_items DROP CONSTRAINT aged_creditor_items_pkey',],
+        #),
```
paulmwatson, line 36: Minor thing but lets delete commented out code. We can always get it back with git if needed.

paulmwatson (approved):

municipal_finance/migrations/0041_create_new_key_columns.py
michaelglenister: Done

municipal_finance/models/aged_creditor.py
```diff
@@ -35,19 +35,20 @@ class Meta:
...
 
 
 class AgedCreditorFactsV1(AgedCreditorFacts):
-    item_code = models.ForeignKey(
+    item = models.ForeignKey(
         AgedCreditorItemsV1,
         models.DO_NOTHING,
-        db_column="item_code",
     )
+    item_code = models.TextField()
```
jbothma, line 51: item_code should get dropped from the facts table. we won't need it any more. We can get it by joining on the dimension table.

Instead of joining the dimension table on item_code https://github.com/OpenUpSA/municipal-data/blob/master/models/aged_creditor.json#L67 we need to change the babbage model to join on the generated column name item_id. I'm not sure you need to actually use an array like this https://github.com/OpenUpSA/municipal-data/blob/master/models/aged_creditor_v2.json#L67 - I think it should detect the related column, but if just changing item_code in the model to item_id doesn't work (facts and aggregate queries return the expected results), then try the way v2 is configured.

municipal_finance/migrations/0042_set_indexes.py
```diff
@@ -0,0 +1,30 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('municipal_finance', '0041_create_new_key_columns'),
+    ]
+
+    operations = [
+        migrations.RunPython(index_keys),
+        migrations.RunPython(drop_fact_item_codes),
```
jbothma, line 29: is this a todo?

municipal_finance/models/aged_creditor.py
michaelglenister: Changed this to use the array, it didn't detect the related column

municipal_finance/migrations/0042_set_indexes.py
michaelglenister: Yup, also ran into a problem dropping item_code in this migration as it's the item_code field is referenced so I added a seperate migration

municipal_finance/resources.py
```diff
@@ -193,7 +193,7 @@ class Meta:
 class AgedCreditorItemsV1Resource(resources.ModelResource):
     class Meta:
         model = AgedCreditorItemsV1
-        import_id_fields = ("code",)
+        import_id_fields = ("code","id")
```
paulmwatson, line 196: Just checking but in `aged_creditor.json` it says `"join_column": ["item_id", "id"],` while here it says `import_id_fields = ("code","id")`. Is `item_id` and `code` right?

paulmwatson (approved):

PR #612, opened 2022-09-01
Demarcation changes fix and add test

michaelglenister:
Issue: https://trello.com/c/gpT3WqUI/518-demarcation-changes-is-broken-again-add-a-test


assets/js/webflow/import.js
```diff
@@ -73,34 +73,13 @@ ga('create', '{{ GOOGLE_ANALYTICS_ID }}', 'auto');\
...
-  {{ notice.content | safe }}\
-</div>\
-{% endfor %}\
-");
-
-$(".profile-notice").html("{% for notice in site_notices %}\
+$(".site-notice").html("{% for notice in site_notices %}\
   <div class='container'>\
-      <div class='profile-notice__text' style='color:black'>{{ notice.content | safe }}</div>\
+    <div class='site-notice__text' style='color:white;margin-top:10px;'>{{ notice.content | safe }}</div>\
```
paulmwatson, line 78: Minor but we should use a CSS class to modify rather than inline styles.

assets/js/webflow/import.js
```diff
@@ -73,34 +73,13 @@ ga('create', '{{ GOOGLE_ANALYTICS_ID }}', 'auto');\
...
 
-$(".section--home-hero .container .layout-grid .layout-grid__col").prepend("{% for notice in site_notices %}\
-<div>\
-  {{ notice.content | safe }}\
-</div>\
-{% endfor %}\
-");
-
-$(".profile-notice").html("{% for notice in site_notices %}\
+$(".site-notice").html("{% for notice in site_notices %}\
```
paulmwatson, line 76: Minor but in future you can use a JavaScript literal to do multiilines rather than having to `\` at the end of each line.  e.g.
```
$(".site-notice").html(`{% for notice in site_notices %}
  <div class='container'>
    <div class='site-notice__text' style='color:white;margin-top:10px;'>{{ notice.content | safe }}</div>
  </div>
{% endfor %}
`);
```

michaelglenister: Changed this section to a JS literal, most of the import script uses slashes but will use literals from now on

paulmwatson (approved):

PR #617, opened 2022-09-26
Set search params in URL

michaelglenister:
Issue: https://trello.com/c/pwl9HUcL/552-search-back-button-as-a-user-i-would-like-to-be-able-to-go-back-to-my-previous-search-using-the-browser-back-button-so-that-i-ca


paulmwatson (approved):

infrastructure/static/js/mm-webflow.js
```diff
@@ -686,7 +697,29 @@ function mmWebflow(js) {
       }
```
paulmwatson, line 697: Blank function?

infrastructure/static/js/mm-webflow.js
```diff
@@ -686,7 +697,29 @@ function mmWebflow(js) {
...
     });
 
-    triggerSearch();
+    function onPopstate(event) {
+      loadSearchStateFromCurrentURL();
+      triggerSearch(listView.search.createUrl(), false);
+    }
+
+    function loadSearchStateFromCurrentURL() {
+      const queryString = window.location.search.substring(1);
```
paulmwatson, line 706: Browsers have a good, reliable URL params method now and we don't need to chop it up ourselves. `URLSearchParams` https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

infrastructure/static/js/mm-webflow.js
michaelglenister: Looks fine to me, maybe Github minimized the code?

infrastructure/static/js/mm-webflow.js
michaelglenister: Changed. All I did was remove this nasty, redundant bit that manually chops up the queryString. The URLSearchParams approach is used inside `listView.loadSearchFromUrl(queryString);`

infrastructure/static/js/mm-webflow.js
paulmwatson: Oh dumb me, thanks.

paulmwatson (approved):

PR #625, opened 2022-10-17
Cube landing page

michaelglenister:
Issue: https://trello.com/c/3kBfz9Lj/570-v1-vs-v2-as-a-user-i-would-like-to-know-what-v1-and-v2-in-the-api-docs-refer-to-so-that-i-can-tell-which-cubes-i-should-use-to-f

### DIFF
--- a/bin/start-web.sh
+++ b/bin/start-web.sh
@@ -4,5 +4,6 @@ set -o errexit
 set -o pipefail
 set -o nounset
 
+
 python manage.py collectstatic --noinput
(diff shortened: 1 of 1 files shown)


michaelglenister:
[start-web.sh](https://github.com/OpenUpSA/municipal-data/pull/625/files#diff-f32a45ad3f52380bffb952e0fba5e70804c3bb92b16e5535a8979511aaac46ce) comes from a change to make Railway deploys work

paulmwatson (approved):

PR #626, opened 2022-10-25
Add data driven summaries to the data portal

michaelglenister:
Issue: https://trello.com/c/2qybjQ9z/604-make-data-summaries-data-driven

### DIFF
--- a/municipal_finance/static/stylesheets/site.scss
+++ b/municipal_finance/static/stylesheets/site.scss
@@ -189,6 +189,12 @@ $black: #333;
 .former-name {
   color: #aeaeae;
 }
+.disestablished-note {
+  font-size: 15px;
+}
+.lead {
+  margin-bottom: 15px;
+}
 @media only screen and (max-width: 991px) {
   .cube {
     margin-left: 0px;
--- a/municipal_finance/templates/index.html
+++ b/municipal_finance/templates/index.html
@@ -26,10 +26,10 @@
         </div>
 
         <div class="col-sm-4" id="municipalities">
-          <h2>{{ summary.municipalities.total }} municipalities</h2>
+          <h2>{{ summary.municipalities.total }} municipalities*</h2>
           <p class="lead">
             {{ summary.municipalities.metros }} metros, {{ summary.municipalities.districts }} district and {{summary.municipalities.munis }} local municipalities.
-            <br><sup>* includes disestablished municipalities</sup>
+            <p class="disestablished-note">* includes disestablished municipalities</p>
           </p>
         </div>
 


paulmwatson (approved):

paulmwatson (approved):

municipal_finance/summarise_data.py
```diff
@@ -0,0 +1,119 @@
...
+def summarise():
+    min_year = 3000
+    max_year = 1000
+    count_facts = 0
+
+    for table in FACT_TABLES:
+        min_year = get_min(min_year, table)
+        max_year = get_max(max_year, table)
+
+    count_years = max_year - min_year
```
jbothma, line 83: is this an off by one error? looks like when I upload a file with a single year of data, it says I have zero years of data

municipal_finance/summarise_data.py
```diff
@@ -0,0 +1,119 @@
...
+    Summary.objects.update_or_create(
+        type='years',
+        defaults={
+            'content': f'{{"count":{count_years}, "min":{min_year}, "max":{max_year}}}'}
+    )
+
+    total = Geography.objects.all().count()
+    metros = Geography.objects.filter(category='A').count()
+    munis = Geography.objects.filter(category='B').count()
+    districts = Geography.objects.filter(category='C').count()
```
jbothma, line 93: this needs to exclude disestablished municipalities https://municipaldata.treasury.gov.za/api/cubes/demarcation_changes/facts

jbothma: otherwise you'll be over-counting

jbothma (changes requested):
great! Just a couple of smallish changes needed I think

Could you also add a test user to demodata and document a la https://github.com/OpenUpSA/wazimap-ng/blob/staging/demodata.json#L145

municipal_finance/summarise_data.py
michaelglenister: Changed to count a list of years and get min max from that without magic numbers

municipal_finance/summarise_data.py
michaelglenister: Discussed in standup - added a note making it clear that disestablished municipalities are included in the data

jbothma: nice - I don't think `sup` is the right thing - I think it needs space below the big text, and a matching *. The size of the text is fine, so just style it another way to get smaller text and a bit more space

michaelglenister: Changed to <p>, added matching asterisk and extra space

jbothma (approved):

PR #629, opened 2022-11-07
Dropdown double click

michaelglenister:
Issue: https://trello.com/c/nG0KhKlh/603-capital-project-dropdowns-have-to-be-clicked-twice-to-open


infrastructure/tests/test_search_page.py
```diff
@@ -298,5 +298,4 @@ def test_search_events(self):
         # Remove filters with a dropdown menu
         self.click("#functions-dropdown")
         self.click("#functions-dropdown")
```
paulmwatson, line 300: Why `self.click("#functions-dropdown")` twice?

michaelglenister: The first click expands the dropdown and the second click selects the first option which is 'All Government Functions' to remove the filter

Just above this, the test expands the dropdown and then selects a specific project function to filter by

paulmwatson (approved):

PR #649, opened 2023-01-17
Set x axis ordering for Plotfy

michaelglenister:
Issue: https://trello.com/c/almeayLB/624-household-bills-years-are-shown-out-of-order

### DIFF
--- a/household/static/household/js/household.js
+++ b/household/static/household/js/household.js
@@ -3,19 +3,28 @@ function amount_convert(value) {
 }
 function overall_chart(container, chartData) {
   var data = [];
+  let xaxis = [];
   for (const [income, value] of Object.entries(chartData)) {
     var region = {
-	    name: income,
-	    type: 'bar',
-	    y: value.y,
-	    x: value.x,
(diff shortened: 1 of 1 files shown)


household/static/household/js/household.js
```diff
@@ -3,19 +3,28 @@ function amount_convert(value) {
...
-	    hoverinfo: 'text',
-	    text: value.y.map(amount_convert),
+      name: income,
+      type: 'bar',
+      y: value.y,
+      x: value.x,
+      hoverinfo: 'text',
+      text: value.y.map(amount_convert),
     };
+    xaxis = _.union(xaxis, value.x);
```
paulmwatson, line 16: @michaelglenister should this be inside the for loop and without a push to the xaxis array? I'm not familiar with the `value.x` which may be an array itself?

paulmwatson, line 16: Ignore previous comment, I see union joining previous value of xaxis to value.x. 👍🏼

paulmwatson (approved):

PR #653, opened 2023-02-06
Feature 625 additional custom event tracking

michaelglenister:
Issue: https://trello.com/c/op92KSWU/625-additional-custom-event-tracking-using-google-analytics

### DIFF
--- a/assets/js/components/contacts.js
+++ b/assets/js/components/contacts.js
@@ -120,18 +120,16 @@ export class ContactSection {
 
     this.$staffLinks = this.$contactContainer.find(".expand-block__trigger a");
     this.$muniLinks = this.$element.find("a");
-    this.$staffLinks.on('click', ((e) => {
-      gtag('event', 'contact_info', {
+    this.callTagEvent(this.$staffLinks, `${$(e.target).text()}`);
+    this.callTagEvent(this.$muniLinks, "Municipal wide contacts");
+  }
+
+  callTagEvent(target, label) {
+    target.on('click', ((e) => {
+      gtag('event', "contact_info", {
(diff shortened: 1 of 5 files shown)


assets/js/components/contacts.js
```diff
@@ -109,6 +118,21 @@ export class ContactSection {
...
+        category: "Contacts",
+        action: "Click",
+        label: `${$(e.target).text()}`,
+      });
+    }));
+    this.$muniLinks.on('click', ((e) => {
+      gtag('event', 'contact_info', {
+        category: "Contacts",
+        action: "Click",
+        label: `Municipal wide contacts`,
```
paulmwatson, line 134: Don't need `` if no variable interpolation.

assets/js/components/contacts.js
```diff
@@ -109,6 +118,21 @@ export class ContactSection {
...
+      gtag('event', 'contact_info', {
+        category: "Contacts",
+        action: "Click",
+        label: `${$(e.target).text()}`,
+      });
+    }));
+    this.$muniLinks.on('click', ((e) => {
+      gtag('event', 'contact_info', {
+        category: "Contacts",
+        action: "Click",
```
paulmwatson, line 133: A bit repetitive, just the target and label different. Not major though.

assets/js/components/contacts.js
michaelglenister: Changed literal to regular string

assets/js/components/contacts.js
michaelglenister: Moved these tags to a function

paulmwatson (approved):

PR #662, opened 2023-02-20
Find and download all videos

michaelglenister:
Issue: https://trello.com/c/QNhASQln/369-all-videos-as-a-curious-user-i-would-like-to-be-able-to-find-all-videos-so-that-i-can-study-municipal-finances-without-having-to


webflow/templates/webflow/index.html
```diff
@@ -575,8 +577,14 @@ <h6 class="small-heading"><span class="search__result-value">0</span> Results</h
                     }
                   </script>
               </a>
+              <a href="#" class="video_download-button is-home w-inline-block">
+                <p class="video_download-button-text">Download this video</p>
+              </a>
             </div>
           </div>
+          <p id="w-node-b2c609bc-8102-5ca8-f0a5-6628f4b42c8d-cc14bbed" class="video_view-more is-space-above">
+            <a href="#">See more videos like this</a>
```
paulmwatson, line 586: Is it a Webflow thing that these are blank targets in code but then a real target is attached at run time? ```$('.video_view-more a').attr('href', '/help#informational-videos');``` @michaelglenister

michaelglenister: I don't think there's a specific reason for it to be added at runtime, if anything it's probably better to be included in Webflow itself

michaelglenister: Imported the Webflow fix for this

PR #677, opened 2023-04-06
Infrastructure projects - obvious searches don't work

michaelglenister:
Issue: https://trello.com/c/mRr8X1gH/549-obvious-searches-dont-work


infrastructure/views/api.py
```diff
@@ -163,7 +163,10 @@ def text_search(self, qs, text):
         if len(text) == 0:
             return qs
 
-        return qs.filter(content_search=SearchQuery(text))
+        result = qs.filter(
+            Q(project_description__contains=text) | Q(content_search=SearchQuery(text))
```
paulmwatson, line 167: Does it matter the search is case sensitive? I tried searching for `dist` as `P-CIN WAT DISTRIB POINT` appears in the test data but it returned no results. I then searched for `DIST` and it returned results.

paulmwatson (changes requested):

infrastructure/views/api.py
michaelglenister: Changing the field lookup to `__icontains` seems to have solved case sensitivity for partial matches

paulmwatson (approved):

PR #692, opened 2023-06-28
Bulk downloads

michaelglenister:
Story: https://trello.com/c/MbTlEBqv/673-bulk-download-as-a-user-i-would-like-to-be-able-to-download-any-and-all-the-in-the-system-so-that-i-can-use-it-for-my-own-analys

In deploy preview: Ensure all environment vars are set (SITE_ID, AWS_S3_ENDPOINT_URL, AWS_STORAGE_BUCKET_NAME)

When deploying to prod: Ensure test data in S3 buckets are cleared out 

For future development: Use a separate directory so as to not interfere with prod data


municipal_finance/settings.py
```diff
@@ -450,30 +450,23 @@
...
-    AWS_STORAGE_BUCKET_NAME = env.str("AWS_STORAGE_BUCKET_NAME")
-    AWS_DEFAULT_ACL = "public-read"
-    AWS_BUCKET_ACL = "public-read"
-    AWS_AUTO_CREATE_BUCKET = True
-    AWS_S3_ENDPOINT_URL = env.str("AWS_S3_ENDPOINT_URL", None)
-    AWS_S3_REGION_NAME = env.str("AWS_S3_REGION_NAME", None)
-    AWS_S3_SECURE_URLS = env.bool("AWS_S3_SECURE_URLS", True)
-    AWS_S3_CUSTOM_DOMAIN = env.str("AWS_S3_CUSTOM_DOMAIN", None)
-    AWS_S3_FILE_OVERWRITE = False
+DEFAULT_FILE_STORAGE = "municipal_finance.storage.MediaStorage"
```
jbothma, line 453: this should be configurable via env var and default to s3BotoStorage so that minio is used in dev

municipal_finance/settings.py
```diff
@@ -450,30 +450,23 @@
...
+    "AWS_STORAGE_BUCKET_NAME", "munimoney-media"
+)
+AWS_DEFAULT_ACL = "public-read"
+AWS_BUCKET_ACL = "public-read"
+AWS_AUTO_CREATE_BUCKET = True
+AWS_S3_ENDPOINT_URL = env.str("AWS_S3_ENDPOINT_URL", None)
+AWS_S3_REGION_NAME = env.str("AWS_S3_REGION_NAME", None)
+AWS_S3_SECURE_URLS = env.bool("AWS_S3_SECURE_URLS", True)
+AWS_S3_CUSTOM_DOMAIN = "%s.s3.amazonaws.com" % AWS_STORAGE_BUCKET_NAME
+AWS_S3_FILE_OVERWRITE = False
```
jbothma, line 466: we didn't used to have this in prod (DEBUG=False) so I don't think we should add it now

municipal_finance/settings.py
michaelglenister: Changed to be an optional env var with the default "municipal_finance.storage.MediaStorage"

paulmwatson (approved):

PR #714, opened 2023-10-03
Set S3_URL_PREFIX to link to bulk downloads

michaelglenister:
Issue: https://trello.com/c/qhVEIWhC/719-django-worker-intermittent-error-when-reading-from-s3

Slightly unusual PR:

- The actual fix for the story: https://github.com/OpenUpSA/municipal-data/blob/bd5ae49326ccc77a88e0faa546f6410e4683c668/municipal_finance/storage.py#L7
- Changes involving S3_URL_PREFIX are to separate the URL that bulk downloads uses from the one S3Boto3 storage uses which is in AWS_S3_ENDPOINT_URL
- Changes adding kwargs to cube update functions are to fix an error caused by not handling the cube_model argumeant added  by this [PR](https://github.com/OpenUpSA/municipal-data/pull/712/files)
- Dependency updates, might not be necessary for the fix but I just left it in anyway


municipal_finance/storage.py
```diff
@@ -4,3 +4,4 @@
 class MediaStorage(S3Boto3Storage):
     location = "media"
     file_overwrite = False
+    region_name = "eu-west-1"
```
paulmwatson, line 7: We should make this an environment variable (with `eu-west-1` as the default.)

PR #718, opened 2023-10-16
Use boto3 to get the bulk download index file

michaelglenister:
Follow up fix for:https://trello.com/c/MbTlEBqv/673-bulk-download-as-a-user-i-would-like-to-be-able-to-download-any-and-all-the-in-the-system-so-that-i-can-use-it-for-my-own-analys


michaelglenister:
I think this was not needed or was added by a different PR

PR #729, opened 2024-01-09
Migration cleanup

michaelglenister:
Story: https://trello.com/c/tQxOY4vq/736-maintenance-i-migrations-into-a-reliable-schema

Follow-up patch: https://github.com/OpenUpSA/municipal-data/pull/736


paulmwatson:
Same error for other app:
```
docker-compose run --rm portal python manage.py migrate
django.db.utils.ProgrammingError: relation "aged_creditor_items" already exists
```

paulmwatson (approved):
Ah yes I see about running a squash as `--fake`. In local I then made a model change, ran `makemigrations`, and then ran `migrate` (without `fake`) and it worked. So all good.

